PR #4, opened 2015-07-05
Update the homepage in the gemspec to point to the right repo

rtyler:


rtyler:
Travis is not going to succeed until the next PR is merged, this is minor an of little consequence so I'm merging ahead of the checkmark

PR #7, opened 2015-08-17
add kafka tracer

jamescway:
Adding Kafka tracer functionality and documentation.  Allows using Kafka as the transport instead of scribe.

Caveat: This implementation can only be used by Jruby (JVM based) and will not be usable by MRI ruby.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,4 +3,5 @@ rvm:
 - 2.1
 - 1.9.3
 - jruby
-- jruby-head
+- jruby-1.7.21
+
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ where Rails.config.zipkin_tracer or config is a hash that can contain the follow
    response status, headers, and body; and can record annotations
  * `:filter_plugin` - plugin function which recieves Rack env and will skip tracing if it returns false
  * `:whitelist_plugin` - plugin function which recieves Rack env and will force sampling if it returns true
+ * `:zookeeper` - plugin function which uses zookeeper and kafka instead of scribe as the transport
 
 ## Warning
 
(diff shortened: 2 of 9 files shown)


.travis.yml
```diff
@@ -3,4 +3,3 @@ rvm:
 - 2.1
 - 1.9.3
 - jruby
-- jruby-head
```
jamescway: @rtyler any reason we want to use head here?  Was making builds fail.

rtyler: It depends what version of RVM is being used in Travis CI. `jruby-head` in older versions refers to JRuby 9k, which has since been released and is the current version of JRuby.

In newer versions of RVM, `jruby` refers to `jruby-9.0.0.0`. We should definitely still be testing on JRuby 1.7.x (1.7.21) and JRuby 9k

lib/zipkin-tracer.rb
```diff
@@ -17,6 +17,15 @@
 
 require 'zipkin-tracer/careless_scribe'
 
+if RUBY_PLATFORM == 'java'
+  begin
+    require 'hermann/producer'
+  rescue LoadError
+    # Hermann not available
+  end
+  require 'zipkin-tracer/zipkin_kafka_tracer' if defined?(::Hermann)
```
rtyler: If Hermann cannot be loaded at this time, will the tracer actually operate correctly? I would expect this to be a failure instead of a recoverable error

README.md
```diff
@@ -77,3 +78,15 @@ For example:
...
     lambda {|env| KNOWN_DEVICES.include?(env['HTTP_X_DEVICE_ID'])}
+
+### Kafka Tracer
+
+Kafka tracer inherits from Tracer which is found in Finagle.  It allows using Kafka as
+the transport instead of scribe.  If a config[:zookeeper] parameter is pass into the
+initialization of the RackHandler, then the gem will use Kafka.  Hermann is the kafka
+client library.  Hermann and the Scribe gems are optionally installed because we would
+only use one of the other.  In your application, you will need to explicitly install
+either Hermann or Scribe.
```
rtyler: IMO adding the specific statements from a `Gemfile` here would be useful, e.g.:


```
  # zipkin-kafka-tracer requires at minimim Hermann 0.25.0 or later
  gem 'hermann', '~> 0.25'
```

```
  # scribe?!
 gem 'somescribe'
```

.travis.yml
jamescway: It looks like it was using jruby-9.0.1.0-SNAPSHOT which seems like is a little more edge than we need?

rtyler: That is correct, I suggest making this:

```
- jruby
- jruby-1.7.21
```

jamescway:
not sure why github is not updating the PR, closing to make a new one

PR #9, opened 2015-09-04
Use finagle-thrift ~> 1.4.1

jfeltesse-mdsol:
Makes the trace handling thread safe and suppresses Hash deprecation warnings.

See also https://github.com/twitter/finagle/commit/086f36280d95ce56e9a30eda789c983ae79b82ec

Bumped to 0.4.0 because finagle-thrift went from 1.3.x to 1.4.x.

Took the opportunity to remove the redundant scribe dependency in the gemspec.

### DIFF
--- a/lib/zipkin-tracer/version.rb
+++ b/lib/zipkin-tracer/version.rb
@@ -12,6 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 module ZipkinTracer
-  VERSION = "0.3.1"
+  VERSION = "0.4.0"
 end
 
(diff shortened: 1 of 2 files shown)


jamescway:
:+1: I guess another nail in the coffin for scribe

jfeltesse-mdsol:
well, scribe isn't removed or anything, it's just that it appeared 2 times in the gemspec and IIRC the first time I tried to bundle that gem in a different repo it complained about it.

jamescway:
icic, thanks for finding that bug

PR #10, opened 2015-09-06
moved faraday middleware into the zipkin client gem

JordiPolo:
Adds the code in the https://github.com/openzipkin/faraday-zipkin gem into this gem.
I think there are very good reasons to do this:
- Although possible in theory, in practice you will not use faraday-zipkin without zipkin-tracer, so they were tied together anyways
- They communicate by the Trace.id variable, if this ever changes in one side (rack middleware) it will be much easier to modify its consumers
- Coherent use of thrift versions, etc. We are using their internal classes so use of different versions may cause weird bugs
- Opportunities to DRY code, write integration tests.

I think after merging this, we can have a PR in https://github.com/openzipkin/faraday-zipkin in the README declaring it deprecated.

Other than bringing the code of Faraday, I've tried to refrain to make any other code changes in this PR.
The other changes are:
- Added info on how to use Faraday in the README and the travis build badge of the project
- Added a Changelog file
-  Added .rspec file (from faraday gem)
- changed the structure of the lib directory, now we have a rack/ and a faraday/ directories
- changed the name of the faraday middleware to be called ZipkinTracer::FaradayHandler, so it is coherent with our already existing ZipkinTracer::RackHandler
- Solved problem of faraday spec randomly failing

@jamescway  please review


JordiPolo:
ping @jamescway  @adriancole

jamescway:
@JordiPolo Thanks for putting thought into this.  I do agree that sharing the similar code is a good thing  A few questions to think through before we consider merging:
* in some use cases there is only a need for one or the other.  Wouldn't that be importing some unnecessary libs in that case?  I will say that using faraday-zipkin only is kind of ugly and needs work.
* a case can be made to standardize on faraday middleware, but this would be forcing that issue, wouldn't this implementation impede someone creating `other_middleware-zipkin`

JordiPolo:
@jamescway  Thanks for the review

I believe 4863bbd addresses your concerns.
Now we do not force the Faraday dependency but allow the faraday middleware to be available if Faraday is available.

Also, I think using the same trick this gem can gather any other middleware or connector needed for Zipkin. Just Faraday probably is what we will ever need though, as is kind of standard.

lib/zipkin-tracer.rb
```diff
@@ -1,113 +1,7 @@
...
-
-      Trace::TraceId.new(*trace_parameters)
-    end
-  end
-
-end
+begin
+  require 'faraday'
+  require 'zipkin-tracer/faraday/zipkin-tracer'
+rescue LoadError #Faraday is not available, we do not load our code.
```
adriancole: cool.. so faraday is optional due to this rescue

adriancole:
looks good! thanks for addressing @jamescway's concern about requiring faraday. keep'em coming

PR #12, opened 2015-09-09
extracting out config

jcarres-mdsol:
Part of a grand plan to spec better the code.
configuration related code and logic lives in its own file and added a spec for it
@jamescway @adriancole


adriancole:
I love changes like this.

PR #13, opened 2015-09-09
Feature/more specs

jcarres-mdsol:
This PR:
- Rescue errors in the rack middleware so if Zipkin somehow blows up, the app still continues its probably more important work
- Adds an integration spec that actually starts up two applications and makes them communicate. It checks that the trace Ids are really sent between them so we know that the rack and faraday middleware are working correctly together.

Also small things like upping version of rspec and adding versions of gems to the spec file. Helps when new people gets into the project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,12 @@ responses).
 
 ### Sending traces on outgoing requests with Faraday
 
-First, for the Faraday middleware to have the correct trace ID, you should be using the rack middleware in your application.
+First, Faraday has to be part of your Gemfile:
+```
+gem 'faraday', '~> 0.8'
+```
+
+For the Faraday middleware to have the correct trace ID, the rack middleware should be used in your application as explained above.
 
 Then include ZipkinTracer::FaradayHandler as a Faraday middleware:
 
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -39,7 +39,12 @@ responses).
 
 ### Sending traces on outgoing requests with Faraday
 
-First, for the Faraday middleware to have the correct trace ID, you should be using the rack middleware in your application.
+First, Faraday has to be part of your Gemfile:
```
jcarres-mdsol: somehow I merge back with master and got this. Not really a change, that text is in master already

lib/zipkin-tracer.rb
```diff
@@ -1,2 +1,7 @@
 require 'zipkin-tracer/rack/zipkin-tracer'
-require 'zipkin-tracer/faraday/zipkin-tracer'
\ No newline at end of file
+
+begin
```
jcarres-mdsol: same than above, this is there already, crazy Github?

adriancole:
I'd squash this into one commit or two commits, latter changelog and version

adriancole:
LGTM otherwise

PR #14, opened 2015-09-10
added a logger option so we can log errors

jcarres-mdsol:
* Added a logger configuration option so we can log errors.
* Log exceptions
* Added an explanation about what will happen if no server information is configured in the gem

@adriancole

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,14 @@ where Rails.config.zipkin_tracer or config is a hash that can contain the follow
  * `:filter_plugin` - plugin function which recieves Rack env and will skip tracing if it returns false
  * `:whitelist_plugin` - plugin function which recieves Rack env and will force sampling if it returns true
  * `:zookeeper` - plugin function which uses zookeeper and kafka instead of scribe as the transport
+ * `:logger` - A logger class following the standard's library Logger interface (Log4r, Rails.logger, etc).
+
+
+ If The configuration do not provide a Scribe or a Zookeeper servers, then the middlewares will not
+ attempt to send traces. But they will still generate proper IDs and pass them to other services.
+ Thus, if you only want to generate IDs for instance for logging and do not intent to integrate with Zipkin,
+ you can still use this gem. Just do not specify any server :)
+
 
 ### Warning
 
--- a/lib/zipkin-tracer/rack/zipkin-tracer.rb
+++ b/lib/zipkin-tracer/rack/zipkin-tracer.rb
@@ -94,7 +94,9 @@ module ZipkinTracer extend self
       @lock.synchronize do
         yield
       end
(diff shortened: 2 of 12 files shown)


README.md
```diff
@@ -27,6 +27,14 @@ where Rails.config.zipkin_tracer or config is a hash that can contain the follow
  * `:filter_plugin` - plugin function which recieves Rack env and will skip tracing if it returns false
  * `:whitelist_plugin` - plugin function which recieves Rack env and will force sampling if it returns true
  * `:zookeeper` - plugin function which uses zookeeper and kafka instead of scribe as the transport
+ * `:logger` - A logger class following the standard's library Logger interface (Log4r, Rails.logger, etc).
+
+
+ If The configuration do not provide a Scribe or a Zookeeper servers, then the middlewares will not
+ attempt to send traces. But they will still generate proper IDs and pass them to other services.
+ Thus, if you only want to generate IDs for instance for logging and do not intent to integrate with Zipkin,
```
adriancole: hehe

spec/lib/config_spec.rb
```diff
@@ -28,6 +29,26 @@ module ZipkinTracer
...
+        logger = 'TrusmisLogger'
+        object_double("Rails", logger: logger).as_stubbed_const
+
+        config = Config.new(nil, {})
+        expect(config.logger).to eq(logger)
+      end
+
+      it 'uses STDOUT if nothing was provided and not using rails' do
+        logger = 'TrusmisLogger'
+       # object_double("Rails", logger: logger).as_stubbed_const
```
adriancole: intended?

lib/zipkin-tracer/rack/zipkin-tracer.rb
```diff
@@ -94,7 +94,8 @@ def synchronize(&block)
       @lock.synchronize do
         yield
       end
-    rescue# Nothing wonky that the tracer does should stop us from using the app!!!
+    rescue StandardError => e # Nothing wonky that the tracer does should stop us from using the app!!!
+      @config.logger.error("Exception #{e.message} while sending Zipkin traces. #{e.backtrace}")
```
adriancole: should this be error level? if tracer shouldn't stop us from using the app, and we have a problem that issues an error, could the amount of errors take down the app (ex. fill-up log files)

JordiPolo:
It fails, I have the feeling jruby is telling me something is wrong. I'll check it tomorrow

lib/zipkin-tracer/rack/zipkin-tracer.rb
JordiPolo: I think it should be error level.
Any app should have log rotation or other system to avoid the log to fill the disk space. If we can bring it down with this, other many gems/pieces or their app would be able to do so also.

jcarres-mdsol:
#15  supersedes this

PR #15, opened 2015-09-11
add logger support and rescue all exceptions. Safety first

jcarres-mdsol:
- Rescue low level errors in both middlewares
- Added a logger configuration option so we can log errors.
- Log exceptions
- Added an explanation about what will happen if no server information is configured in the gem

@adriancole  @jamescway

### DIFF
--- a/lib/zipkin-tracer/faraday/zipkin-tracer.rb
+++ b/lib/zipkin-tracer/faraday/zipkin-tracer.rb
@@ -33,15 +33,23 @@ module ZipkinTracer
 
         # annotate with method (GET/POST/etc.) and uri path
         ::Trace.set_rpc_name(env[:method].to_s.upcase)
-        ::Trace.record(::Trace::BinaryAnnotation.new("http.uri", url.path, "STRING", endpoint))
-        ::Trace.record(::Trace::Annotation.new(::Trace::Annotation::CLIENT_SEND, endpoint))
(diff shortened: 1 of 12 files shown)


lib/zipkin-tracer/faraday/zipkin-tracer.rb
```diff
@@ -33,17 +33,25 @@ def call(env)
...
         result
       end
     end
 
+    private
+
+    def record(annotation)
+      ::Trace.record(annotation)
+    rescue Exception # Sockets errors inherit from Exception, not from StandardError
+      #TODO: if this class some day accepts a config hash, add a logger
```
jamescway: make an issue for this?

spec/integration/integration_spec.rb
```diff
@@ -15,9 +15,9 @@
     @port2 = 4445
     @base_url2 = "http://localhost:#{@port2}"
     @pipe2 = IO.popen("rackup #{ru_location} -p #{@port2}")
-    sleep(1)
+    sleep(2)
     if RUBY_PLATFORM == 'java'
-      sleep(10) #Jruby starts slow
+      sleep(20) #Jruby starts slow
```
jamescway: do you think it would be worth renaming to `jruby_startup_waittime = 20` (optional)

jamescway:
no blocking comments +1

PR #18, opened 2015-09-24
make hermann a dependency

jamescway:
Removing scribe as an explicit dependency.  In the readme, it states that the user should explicitly install their transport of choice.


phrinx:
yes, making both dependencies optional makes more sense to me

:shipit:

PR #19, opened 2015-10-07
only trace valid requests

jcarres-mdsol:
Realized while looking at the traces that we had bunch of traces for weird URLs.
Turns out is people trying to find security holes in our apps.
We do not want to send those all traces to our zipkin. This PR solves the problem for Rails apps.

@jamescway @jfeltesse-mdsol

### DIFF
--- a/lib/zipkin-tracer/rack/zipkin-tracer.rb
+++ b/lib/zipkin-tracer/rack/zipkin-tracer.rb
@@ -49,7 +49,7 @@ module ZipkinTracer extend self
 
     def call(env)
       # skip certain requests
-      return @app.call(env) if filtered?(env) || !valid_request?(env)
+      return @app.call(env) if filtered?(env) || !routable_request?(env)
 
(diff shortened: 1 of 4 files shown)


lib/zipkin-tracer/rack/zipkin-tracer.rb
```diff
@@ -59,6 +59,16 @@ def call(env)
     end
 
     private
+
+    # If the request is not valid for this service, we do not what to trace it.
+    def valid_request?(env)
```
jfeltesse-mdsol: Can the naming be improved? `valid_` is too broad IMHO. `processable_` or something?

jcarres-mdsol: routable?

jfeltesse-mdsol: `acceptable_`, `routable_`, `processable_`, `something completely different`, pick one :wink:

jamescway:
:+1:

jamescway:
released 0.6.2

PR #20, opened 2015-10-20
faraday should pop the stack properly

jcarres-mdsol:
@jamescway 
Faraday was not properly restoring the spanID after it created a new one in a faraday outbound connection. 
This solves that bug


jamescway:
:+1:

PR #21, opened 2015-10-21
tracing does not wait for ack and connection moved to a different thread

jcarres-mdsol:
Big PR but had to be done this way.

This solves a big issue with the connection between the client and the collector.
The client was waiting for the collector to send back ack saying the spans were saved, that was too slow to the point most spans were timing out.

Now we do not wait. Also we connect in a different thread so we do not slow down the main request.

Also added some new specs and moved code to initializers instead of calling every time if possible.

@jamescway

### DIFF
--- a/lib/zipkin-tracer/careless_scribe.rb
+++ b/lib/zipkin-tracer/careless_scribe.rb
@@ -1,35 +1,70 @@
 # Copyright 2012 Twitter Inc.
-# 
+#
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
-# 
+#
 #      http://www.apache.org/licenses/LICENSE-2.0
-# 
+#
 # Unless required by applicable law or agreed to in writing, software
 # distributed under the License is distributed on an "AS IS" BASIS,
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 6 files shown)


lib/zipkin-tracer/careless_scribe.rb
```diff
@@ -1,35 +1,70 @@
...
 # limitations under the License.
 
-class CarelessScribe
-  def initialize(scribe)
-    @scribe = scribe
+require 'sucker_punch'
+
+
+module ScribeThrift
+  # This is here just for the monkey partching
```
jamescway: s/partching/patching/

lib/zipkin-tracer/careless_scribe.rb
```diff
@@ -1,35 +1,70 @@
...
+  class Client
+    # This method in the original class was both sending and receiving logs.
+    # The original class: https://github.com/twitter/scribe/blob/master/vendor/gen-rb/scribe.rb
+    # Receiving logs may take even several seconds, depending on the buffering of the collector.
+    # We are just sending and forgetting here, we do not really care about the result
+    def Log(messages)
+      send_Log(messages)
+      0 # 0 means success , the original code called recv_Log()
+    end
```
jamescway: IC, thanks for making this very clear

jamescway:
:+1: looks good, I had been thinking about this for a while glad you did it.  Release will be pushed right away.

PR #25, opened 2015-11-06
rescue errors when lookup of the host fails

jcarres-mdsol:
This error happens when running the code inside a VPN. I'm not big on networks but I've seen it myself and some other people also. The error happens using Socket.gethostname. We are rescuing that now. (before it was happening inside default_endpoint)
@jamescway @adriancole


adriancole:
LGTM

PR #27, opened 2015-11-09
downcasing rpc names

jcarres-mdsol:
let lowercase rule the world
@adriancole  @jamescway


adriancole:
while you're at it, can you set endpoint.service_name to lowercase? Can be a separate PR. Thx!

PR #28, opened 2015-11-21
Sa annotated

jcarres-mdsol:
Previously we were sending CS/CR with the name of the service we were calling.
Turns out this is incorrect.
Now we do CS/CR with the local server and we are sending an extra annotation (SA) with the name of the service we are callling
@jamescway  @adriancole


adriancole:
:shipit: thanks tons!

PR #29, opened 2015-12-08
Set caller service name using domain environment variable

ssteeg-mdsol:
The caller service name is determined from the DOMAIN environment variable. Otherwise, it falls back to the default service name defined in the configuration file.

I also found that the specs in `/spec/lib/faraday/zipkin-tracer_spec.rb` file were not behaving properly. The checks inside the `expect(::Trace)` blocks were not actually being evaluated. Instead, they were silently failing and the specs would carry on and pass. I've changed them to use `expect(middleware)` instead.

Also, some minor updates to syntax, since I was touching the files anyways.

@jcarres-mdsol @jfeltesse-mdsol

### DIFF
--- a/lib/zipkin-tracer/rack/zipkin-tracer.rb
+++ b/lib/zipkin-tracer/rack/zipkin-tracer.rb
@@ -29,7 +29,7 @@ module ZipkinTracer extend self
     B3_REQUIRED_HEADERS = %w[HTTP_X_B3_TRACEID HTTP_X_B3_PARENTSPANID HTTP_X_B3_SPANID HTTP_X_B3_SAMPLED]
     B3_OPT_HEADERS = %w[HTTP_X_B3_FLAGS]
 
-    def initialize(app, config=nil)
+    def initialize(app, config = nil)
       @app = app
       @lock = Mutex.new
 
@@ -44,7 +44,8 @@ module ZipkinTracer extend self
       else
         ::Trace::NullTracer.new
       end
-      ::Trace.default_endpoint = ::Trace::Endpoint.new( local_ip, config.service_port , config.service_name)
+
+      ::Trace.default_endpoint = ::Trace::Endpoint.new(local_ip, config.service_port, service_name(config.service_name))
       ::Trace.sample_rate=(config.sample_rate)
 
       @config = config
@@ -53,6 +54,7 @@ module ZipkinTracer extend self
     def call(env)
       # skip certain requests
       return @app.call(env) if filtered?(env) || !routable_request?(env)
+
       whitelisted = force_sample?(env)
       id = get_or_create_trace_id(env, whitelisted) # note that this depends on the sample rate being set
       tracing_filter(id, env, whitelisted) { @app.call(env) }
@@ -60,10 +62,15 @@ module ZipkinTracer extend self
 
     private
 
+    # Use the Domain environment variable to extract the service name, otherwise use the default config name
+    def service_name(default_name)
+      ENV["DOMAIN"].to_s.empty? ? default_name : ENV["DOMAIN"].split('.').first
+    end
+
     # If the request is not valid for this service, we do not what to trace it.
     def routable_request?(env)
       return true unless defined?(Rails) #If not running on a Rails app, we can't verify if it is invalid
-      Rails.application.routes.recognize_path(env["PATH_INFO"])
+      Rails.application.routes.recognize_path(env['PATH_INFO'])
       true
     rescue ActionController::RoutingError
       false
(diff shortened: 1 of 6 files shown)


lib/zipkin-tracer/rack/zipkin-tracer.rb
```diff
@@ -103,7 +111,7 @@ def tracing_filter(trace_id, env, whitelisted=false)
 
     def add_request_information(env)
       ::Trace.set_rpc_name(env["REQUEST_METHOD"].to_s.downcase) # get/post and all that jazz
-      ::Trace.record(::Trace::BinaryAnnotation.new("http.uri", env["PATH_INFO"], "STRING", ::Trace.default_endpoint))
+      ::Trace.record(::Trace::BinaryAnnotation.new('http.uri', env["PATH_INFO"], 'STRING', ::Trace.default_endpoint))
```
jfeltesse-mdsol: double quotes still in there :trollface:

jcarres-mdsol:
Would need changes in the changelog and version files

spec/lib/faraday/zipkin-tracer_spec.rb
```diff
@@ -17,29 +17,29 @@ def on_complete
   end
 
   # Ruby 1.8 didn't support \h
-  HEX_REGEX = RUBY_VERSION >= "1.9.2" ? '\h' : '[0-9a-fA-F]'
+  HEX_REGEX = RUBY_VERSION >= '1.9.2' ? '\h' : '[0-9a-fA-F]'
```
jfeltesse-mdsol: since the syntax uses 1.9 hash syntax this can be removed altogether in favor of say
```ruby
HEX_REGEX = /\A\h{16}\z/
```
(always found it odd to have the var called *_regex but it being actually a string)

lib/zipkin-tracer/rack/zipkin-tracer.rb
```diff
@@ -60,6 +62,12 @@ def call(env)
 
     private
 
+    # Use the Domain environment variable to extract the service name, otherwise use the default config name
+    def service_name(default_name)
+      domain = ENV["DOMAIN"] || ''
```
jcarres-mdsol: alternatively you could do "to_s", nil.to_s becomes an empty string

spec/lib/faraday/zipkin-tracer_spec.rb
```diff
@@ -17,29 +17,29 @@ def on_complete
...
 
-  let(:response_env) { { :status => 200 } }
-  let(:wrapped_app) { lambda{|env| ResponseObject.new(env, response_env)} }
+  let(:response_env) { { status: 200 } }
+  let(:wrapped_app) { lambda{ |env| ResponseObject.new(env, response_env) } }
```
jfeltesse-mdsol: what about using the stabby syntax (from 1.9.2)?

`-> (env) { ResponseObject.new(env, response_env) }`

spec/lib/faraday/zipkin-tracer_spec.rb
```diff
@@ -17,29 +17,29 @@ def on_complete
...
-      :request_headers => Faraday::Utils::Headers.new(headers),
+      method: :post,
+      url: url,
+      body: body,
+      request_headers: Faraday::Utils::Headers.new(headers),
     }
     middleware.call(env)
   end
 
   before(:each) {
```
jfeltesse-mdsol: useless `:each`

spec/lib/rack/zipkin-tracer_spec.rb
```diff
@@ -73,11 +73,33 @@ def mock_env(path = '/', params = {})
...
+
       it 'sets the sample rate' do
         expect(::Trace).to receive(:sample_rate=).with(sample_rate)
         subject.call(mock_env)
       end
+    end
 
+    context 'no domain environment variable' do
+      before do
+        ENV["DOMAIN"] = ''
```
jfeltesse-mdsol: is it ok to change the *real* env during specs? can't this be mocked instead?

lib/zipkin-tracer/rack/zipkin-tracer.rb
ssteeg-mdsol: Good idea, I'll switch it

spec/lib/faraday/zipkin-tracer_spec.rb
ssteeg-mdsol: Hmm, yea it's a bit strange. Thanks for the idea though - I'll keep the naming as is, but switch to the 1.9 syntax

spec/lib/faraday/zipkin-tracer_spec.rb
ssteeg-mdsol: :+1:

jcarres-mdsol:
@jamescway  @adriancole  This should be ready to merge

adriancole:
Nice. Thanks for the help and I the close review

PR #31, opened 2015-12-10
Add a JSON tracer (ZipkinJsonTracer)

jfeltesse-mdsol:
This adds a JSON tracer that sends traces in JSON over HTTP.

Noteworthy bits:
- the ip of the endpoint needs to be a regular string in standard ipv4 notation
- doing the actual JSON serializing in the separate thread so that the host application is not slowed down.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,10 +31,13 @@ where Rails.config.zipkin_tracer or config is a hash that can contain the follow
  * `:whitelist_plugin` - plugin function which recieves Rack env and will force sampling if it returns true
  * `:zookeeper` - plugin function which uses zookeeper and kafka instead of scribe as the transport
  * `:logger` - A logger class following the standard's library Logger interface (Log4r, Rails.logger, etc).
+ * `:json_api_host` - hostname with protocol of a zipkin api instance (e.g. `https://zipkin.example.com`) to use JSON over HTTP instead of Scribe or Kafka
+ * `:traces_buffer` (default: 100) - the number of annotations stored until automatic flush
+   (note that annotations are also flushed when the request is complete)
 
 
- If The configuration do not provide a Scribe or a Zookeeper servers, then the middlewares will not
- attempt to send traces. But they will still generate proper IDs and pass them to other services.
+ If the configuration does not provide either a JSON, Scribe or Zookeeper server then the middlewares will not
+ attempt to send traces although they will still generate proper IDs and pass them to other services.
  Thus, if you only want to generate IDs for instance for logging and do not intent to integrate with Zipkin,
  you can still use this gem. Just do not specify any server :)
 
@@ -141,3 +144,11 @@ based broker discovery being JVM based.
 # Install scribe
   gem 'scribe', "~> 0.2.4"
 ```
+
+### JSON Tracer
+
+Like the other tracers, it inherits from the Tracer found in Finagle.
+If the configuration specifies a `json_api_host` then the gem will serialize the traces to JSON
+and POST them to the defined host (the path to which the data is posted is `/api/v1/spans`).
+If a `traces_buffer` value is provided it will buffer the traces call so as not to hammer your zipkin
+instance. The default is `10`.
--- a/lib/zipkin-tracer/rack/zipkin-tracer.rb
+++ b/lib/zipkin-tracer/rack/zipkin-tracer.rb
@@ -17,11 +17,13 @@ require 'scribe'
 
 require 'zipkin-tracer/config'
 require 'zipkin-tracer/careless_scribe'
+require 'zipkin-tracer/zipkin_json_tracer'
 
 if RUBY_PLATFORM == 'java'
   require 'hermann/producer'
   require 'zipkin-tracer/zipkin_kafka_tracer'
 end
+
 
 module ZipkinTracer extend self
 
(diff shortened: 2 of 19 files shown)


jfeltesse-mdsol:
@jcarres-mdsol @ssteeg-mdsol @adriancole @jamescway please have a look when you have time :eyes:

README.md
```diff
@@ -141,3 +144,11 @@ based broker discovery being JVM based.
 # Install scribe
   gem 'scribe', "~> 0.2.4"
 ```
+
+### JSON Tracer
+
+Like the other tracers it inherits from the Tracer found in Finagle.
```
ssteeg-mdsol: [nit] "Like the other tracers, it inherits [...]"

lib/zipkin-tracer/rack/zipkin-tracer.rb
```diff
@@ -34,19 +36,29 @@ def initialize(app, config = nil)
...
+          ::Trace::NullTracer.new
       end
 
-      ::Trace.default_endpoint = ::Trace::Endpoint.new(local_ip, config.service_port, service_name(config.service_name))
-      ::Trace.sample_rate=(config.sample_rate)
+      ::Trace.default_endpoint = ::Trace::Endpoint.make_endpoint(
+        nil, # auto detect hostname
+        config.service_port,
+        service_name(config.service_name),
+        config.adapter == :json ? :string : :i32
```
ssteeg-mdsol: Personally, I'd prefer this logic be outside the parameter context

ssteeg-mdsol:
Looks good to me. Bit concerning about the failing checks though... Would be nice to have them resolved for peace of mind

zipkin-tracer.gemspec
```diff
@@ -18,30 +18,34 @@ $:.unshift lib unless $:.include?(lib)
...
-  s.required_rubygems_version = ">= 1.3.5"
+  s.required_rubygems_version = '>= 1.3.5'
 
-  s.files                     = Dir.glob("{bin,lib}/**/*")
+  s.files                     = Dir.glob('{bin,lib}/**/*')
   s.require_path              = 'lib'
 
-  s.add_dependency "finagle-thrift", "~> 1.4.1"
-  s.add_dependency "rack", "~> 1.3"
+  s.add_dependency 'thrift'
```
jcarres-mdsol: Is this a hard dependency now? Probably we can leave this out for peope who wants to only use json

README.md
```diff
@@ -31,10 +31,13 @@ where Rails.config.zipkin_tracer or config is a hash that can contain the follow
  * `:whitelist_plugin` - plugin function which recieves Rack env and will force sampling if it returns true
  * `:zookeeper` - plugin function which uses zookeeper and kafka instead of scribe as the transport
  * `:logger` - A logger class following the standard's library Logger interface (Log4r, Rails.logger, etc).
+ * `json_api_host` - hostname with protocol of a zipkin api instance (e.g. `https://zipkin.example.com`) to use JSON over HTTP instead of Scribe or Kafka
```
adriancole: intentionally not : prefixed? Personally, I don't care, except we might as well change the others if we don't want to prefix anymore.

README.md
```diff
@@ -31,10 +31,13 @@ where Rails.config.zipkin_tracer or config is a hash that can contain the follow
...
+ * `traces_buffer` (default: 10) - the number of annotations stored until automatic flush
```
adriancole: Could this number be a bit too small? partial spans are kindof a pain. Ex. in brave, it is 10 spans as opposed to 10 annotations. 

In finagle, it is much nicer, as it goes based on the size of the data 1MB. Since this is json text, we might be able to flush based on size sooner or later.

lib/zipkin-tracer/trace.rb
```diff
@@ -0,0 +1,67 @@
...
+require 'finagle-thrift/trace'
+
+module Trace
+  class Span
+    def to_h
+      {
+        name: @name,
+        traceId: @span_id.trace_id.to_s,
+        id: @span_id.span_id.to_s,
+        parentId: @span_id.parent_id.nil? ? nil : @span_id.parent_id.to_s,
```
adriancole: there's also :timestamp (microseconds, usually first annotation) and :duration (also microseconds, can be derived from last timestamp - first timestamp). A "complete" span should send these two keys.

adriancole:
looks good. Made some notes which are more polish.

One thing I'd like to see is the README updated to prefer the json api.

Ex. would be awesome to add step-by-step, from gem install to actually sending a trace manually via a ruby repl!

lib/zipkin-tracer/trace.rb
jfeltesse-mdsol: saw that in the json api responses when doing GETs but I assumed it was something the server computed to make things easier for the front end side.
none of the tracers do that and it works without sending these so how about it keeping it as is, at least for now?
I'm open to change this if there's a compelling argument in favor of doing this in the client VS the server.

README.md
jfeltesse-mdsol: Actually we run our clients with a buffer of `1000`. I stuck to the `scribe_max_buffer` default here, how does `100` sound then?
Note that eventually the scribe tracer is going to be phased out (isn't it?), that is the reason I added this new configuration key. With a broader name it could be used in the kafka tracer too, if it ever implements some kind of buffering.

README.md
jfeltesse-mdsol: :scream: :scream: :scream: 

will fix.

lib/zipkin-tracer/trace.rb
adriancole: The server is only doing this for portability with legacy instrumentation.
It will eventually stop. In general instrumentation should be sending
complete spans, which can trivially perform this step, and must when they
get to the point of local spans.

README.md
adriancole: 100 sounds less likely to result in incomplete spans, but yeah just keep in
mind this isn't a great attribute as ex spans are the unit of work in
zipkin, not attributes.

lib/zipkin-tracer/trace.rb
adriancole: TL;DR of it is that this is a new tracer, might as well do it right as
opposed to laying little bombs for someone to do later. This
particular thing isn't much effort to complete vs the act of
remembering to opening another issue and fix it with a PR later.

jfeltesse-mdsol: ok, I'll see what I can do. I'll be AFAIK until the end of the week though, this is going to be for next Monday.

jfeltesse-mdsol:
> One thing I'd like to see is the README updated to prefer the json api.

Sounds good.

> Ex. would be awesome to add step-by-step, from gem install to actually sending a trace manually via a ruby repl!

Might not have the time now and this PR is unrelated anyway so allow me to skip this for now.

jfeltesse-mdsol:
force-pushed to address the small comments here and there while keeping log pollution to a minimum.

adriancole:
PS merge when ready! or I can

jcarres-mdsol:
@adriancole there is a small problem with the specs, when that's solved, can be merged

jfeltesse-mdsol:
Fixed the specs (stupid timezone issue), removed ruby 2.x syntax stuff and removed `1.9.3` from the travis file since it's reached EOL in Feb 2015.

adriancole:
Many thanks, all!

PR #32, opened 2015-12-14
Refresh the README

jfeltesse-mdsol:
- make JSON the preferred tracer
- fix typos
- syntax highlight the code blocks
- make the copywriting better

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,47 +8,31 @@ Rack and Faraday integration middlewares for Zipkin tracing.
 
 ### Sending traces on incoming requests
 
-Options can be provided via Rails.config for a Rails 3+ app, or can be passed
-as a hash argument to the Rack plugin.
+Options can be provided via Rails.config for a Rails 3+ app, or can be passed as a hash argument to the Rack plugin.
 
-    require 'zipkin-tracer'
-    use ZipkinTracer::RackHandler, config # config is optional
+```ruby
+require 'zipkin-tracer'
+use ZipkinTracer::RackHandler, config # config is optional
+```
 
-where Rails.config.zipkin_tracer or config is a hash that can contain the following keys:
+where `Rails.config.zipkin_tracer` or `config` is a hash that can contain the following keys:
 
- * `:service_name` (REQUIRED) - the name of the service being traced. There are two ways to configure this
-   value. Either write the service name in the config file or set the "DOMAIN"
-   environment variable (e.g. 'test-service.example.com' or 'test-service'). The
-   environment variable takes precedence over the config file value.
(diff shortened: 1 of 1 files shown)


jfeltesse-mdsol:
@jcarres-mdsol @ssteeg-mdsol @adriancole @jamescway

README.md
```diff
@@ -60,23 +44,54 @@ gem 'faraday', '~> 0.8'
...
+  # default Faraday stack
+  faraday.request :url_encoded
+  faraday.adapter Faraday.default_adapter
+end
+```
+
+Note that supplying the service name for the destination service is optional;
+the tracing will default to a service name derived from the first section of the destination URL (e.g. 'service.example.com' => 'service').
+
+## Tracers
```
ssteeg-mdsol: Maybe add something at the top explaining that you need to choose between one of the three following

README.md
```diff
@@ -8,47 +8,31 @@ Rack and Faraday integration middlewares for Zipkin tracing.
...
+use ZipkinTracer::RackHandler, config # config is optional
+```
 
- If the configuration does not provide either a JSON, Scribe or Zookeeper server then the middlewares will not
- attempt to send traces although they will still generate proper IDs and pass them to other services.
- Thus, if you only want to generate IDs for instance for logging and do not intent to integrate with Zipkin,
- you can still use this gem. Just do not specify any server :)
+where `Rails.config.zipkin_tracer` or `config` is a hash that can contain the following keys:
 
+* `:service_name` **REQUIRED** - the name of the service being traced. There are two ways to configure this value. Either write the service name in the config file or set the "DOMAIN" environment variable (e.g. 'test-service.example.com' or 'test-service'). The environment variable takes precedence over the config file value.
```
ssteeg-mdsol: Maybe overkill, but I feel like this information would be better in a table

ssteeg-mdsol:
:+1:

README.md
jfeltesse-mdsol: There's a note at line 34 stating you have a fourth option: not using any! :santa:

README.md
jfeltesse-mdsol: If only making tables in MD wasn't such a PITA and I wasn't so lazy on a Monday morning... :sweat_smile:

adriancole: Lol

adriancole:
:shipit:

jcarres-mdsol:
Obviously a README has not broken the specs but travis is not passing, can your force push some change to make it run again?

adriancole:
no worries. I'll merge

PR #33, opened 2015-12-15
Make Scribe actually an optional dependency

jfeltesse-mdsol:
Up until now even if you used a different tracer (Kafka, JSON) you still needed to depend on Scribe because of the hard coded require at the top of the rack handler file.
This commit moves the require directives to the appropriate places so that we require only what we need.

@jcarres-mdsol @adriancole @ssteeg-mdsol @jamescway

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 0.9.1
+Make Scribe actually optional by inspecting the conf first and requiring after.
+
 # 0.9.0
 Add a JSON tracer (ZipkinJsonTracer).
 
--- a/README.md
+++ b/README.md
@@ -31,10 +31,6 @@ where `Rails.config.zipkin_tracer` or `config` is a hash that can contain the fo
 * `:whitelist_plugin` - plugin function which receives the Rack env and will force sampling if it returns true
 
 
-If the configuration does not provide either a JSON, Zookeeper or Scribe server then the middlewares will not attempt to send traces although they will still generate proper IDs and pass them to other services.
-Thus, if you only want to generate IDs for instance for logging and do not intent to integrate with Zipkin you can still use this gem. Just do not specify any server :)
-
-
 ### Sending traces on outgoing requests with Faraday
 
 First, Faraday has to be part of your Gemfile:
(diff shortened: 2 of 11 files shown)


jcarres-mdsol:
Looks good, maybe update the Readme if it does not say it already how to use these two.
Also changelog and version . I think it's worth a new minor revision

adriancole:
Love the idea. Merge when ready guys

jcarres-mdsol:
What's going on jruby failing for finagle?

jfeltesse-mdsol:
> What's going on jruby failing for finagle?

I'm on it, unsure if twitter is the right way to call them out though.
https://twitter.com/robotvert/status/676581119756668930

Tested locally on jruby 9.0.4.0, specs pass.

jfeltesse-mdsol:
@jcarres-mdsol the README already mentions that to use scribe or kafka you need to include the gem yourself. I took the opportunity to move the block talking about tracing without calling zipkin into its own section within "Tracers" as it's a type of tracer actually.

jfeltesse-mdsol:
The CI issue seems to be related to bundler 1.11.0 and I can't find a way to force install the previous version in the travis.yml file. Stupid travis has 1.11, doing uninstall beforehand removes some unrelated super old version and when times come to bundle it uses the latest...
I'll try to find a way to hack this through but it might turn out to be impossible...

https://github.com/bundler/bundler/issues/4150#issuecomment-164617910

jfeltesse-mdsol:
Gave up on trying to make travis work properly. That thing is buggy, the doc proudly says you can install and use a different bundler version but nope, with jruby it will always use 1.11 in the end which is probably why it doesn't bundle the gems declared as dependencies. 
How the bundler team let this bug in to begin with beats me...

adriancole:
@rtyler any jruby bundler insight to share?

jcarres-mdsol:
I hope next version of bundler will solve automatically this issue so merging for now

jfeltesse-mdsol:
known bug with workaround, I'll submit a PR with the fix in the travis file

https://github.com/travis-ci/travis-ci/issues/5290#issuecomment-164600246

PR #34, opened 2015-12-16
Force bundler 1.10.6 in TravisCI

jfeltesse-mdsol:
Version 1.11 has a bug preventing gems to be properly installed with jruby

Follow up from PR #33 

See https://github.com/travis-ci/travis-ci/issues/5290


jfeltesse-mdsol:
@jcarres-mdsol @adriancole

adriancole:
good find!

PR #35, opened 2015-12-16
DRY tracer classes

jcarres-mdsol:
@jfeltesse-mdsol  Tell me what you think

Moved jsontracer and kafkatracer to inherit from tracerbase. Makes things a tad simpler and save some duplication there. Deleted the connect method from Kafka which was no more than a weird initializer.
Also the integration spec uses the json configuration so we test that instead of scribe

I have not updated the changelog because this is just a refactor.

### DIFF
--- a/lib/zipkin-tracer/zipkin_json_tracer.rb
+++ b/lib/zipkin-tracer/zipkin_json_tracer.rb
@@ -1,7 +1,6 @@
 require 'json'
-require 'faraday'
-require 'sucker_punch'
-require 'finagle-thrift/tracer'
+require 'zipkin-tracer/zipkin_tracer_base'
+
 
 class AsyncJsonApiClient
   include SuckerPunch::Job
@@ -19,55 +18,9 @@ class AsyncJsonApiClient
 end
 
 module Trace
-  class ZipkinJsonTracer < Tracer
-    TRACER_CATEGORY = "zipkin".freeze
-
-    def initialize(json_api_host, traces_buffer)
-      @json_api_host = json_api_host
-      @traces_buffer = traces_buffer
-      reset
-    end
-
-    def record(id, annotation)
-      return unless id.sampled?
-      span = get_span_for_id(id)
-
-      case annotation
-      when BinaryAnnotation
-        span.binary_annotations << annotation
-      when Annotation
-        span.annotations << annotation
-      end
-
-      @count += 1
-      if @count >= @traces_buffer || (annotation.is_a?(Annotation) && annotation.value == Annotation::SERVER_SEND)
-        flush!
-      end
-    end
-
-    def set_rpc_name(id, name)
-      return unless id.sampled?
-      span = get_span_for_id(id)
-      span.name = name.to_s
-    end
-
-    private
-
-    def get_span_for_id(id)
-      key = id.span_id.to_s
-      @spans[key] ||= begin
-        Span.new("", id)
-      end
-    end
-
-    def reset
-      @count = 0
-      @spans = {}
-    end
-
+  class ZipkinJsonTracer < ZipkinTracerBase
     def flush!
-      AsyncJsonApiClient.new.async.perform(@json_api_host, @spans.values.dup)
-      reset
+      AsyncJsonApiClient.new.async.perform(@options[:json_api_host], @spans.values.dup)
     end
   end
 end
--- a/lib/zipkin-tracer/zipkin_json_tracer.rb
+++ b/lib/zipkin-tracer/zipkin_json_tracer.rb
@@ -18,6 +18,8 @@ class AsyncJsonApiClient
 end
 
 module Trace
+  # This class sends information to the Zipkin API.
+  # The API accepts a JSON representation of a list of spans
   class ZipkinJsonTracer < ZipkinTracerBase
     def flush!
       AsyncJsonApiClient.new.async.perform(@options[:json_api_host], @spans.values.dup)
--- a/lib/zipkin-tracer/zipkin_kafka_tracer.rb
+++ b/lib/zipkin-tracer/zipkin_kafka_tracer.rb
@@ -2,17 +2,11 @@ require 'finagle-thrift'
 require 'finagle-thrift/tracer'
 require 'hermann/producer'
 require 'hermann/discovery/zookeeper'
+require 'zipkin-tracer/zipkin_tracer_base'
 
 module Trace
-  class ZipkinKafkaTracer < Tracer
-    TRACER_CATEGORY     = "zipkin".freeze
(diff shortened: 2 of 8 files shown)


lib/zipkin-tracer/zipkin_tracer_base.rb
```diff
@@ -0,0 +1,62 @@
...
+  # This class is a base for tracers sending information to Zipkin.
+  # It knows about zipkin types of annotations and send traces when the server
+  # is done with its request
+  # Traces dealing with zipkin should inherit from this class and implement the
+  # flush! method which actually sends the information
+  class ZipkinTracerBase < Tracer
+    TRACER_CATEGORY = "zipkin".freeze
+
+    def initialize(options={})
+      @traces_buffer = options[:traces_buffer] || raise(ArgumentError, 'A proper buffer must be setup for the Zipkin tracer')
```
jfeltesse-mdsol: [nits] if you run rubocop on this file it's going to say "use fail instead of raise"

lib/zipkin-tracer/zipkin_tracer_base.rb
```diff
@@ -0,0 +1,62 @@
...
+
+      case annotation
+      when BinaryAnnotation
+        span.binary_annotations << annotation
+      when Annotation
+        span.annotations << annotation
+      end
+
+      @count += 1
+      if @count >= @traces_buffer || (annotation.is_a?(Annotation) && annotation.value == Annotation::SERVER_SEND)
```
jfeltesse-mdsol: kafka doesn't do any buffering so it the buffer should always be "1".
that should probably be hardcoded in its initializer.

lib/zipkin-tracer/zipkin_tracer_base.rb
```diff
@@ -0,0 +1,62 @@
...
+    end
+
+    def set_rpc_name(id, name)
+      return unless id.sampled?
+      span = get_span_for_id(id)
+      span.name = name.to_s
+    end
+
+    def flush!
+      raise 'implement this method in inherit classes'
```
jfeltesse-mdsol: finagle-thrift uses "not implemented". might be preferable just for the sake of consistency.

https://github.com/twitter/finagle/blob/develop/finagle-thrift/src/main/ruby/lib/finagle-thrift/tracer.rb

lib/zipkin-tracer/zipkin_tracer_base.rb
```diff
@@ -0,0 +1,62 @@
...
+    end
+
+    private
+
+    def get_span_for_id(id)
+      key = id.span_id.to_s
+      @spans[key] ||= begin
+        Span.new("", id)
+      end
```
jfeltesse-mdsol: if you're going to do refactorings might as well change this, I never really understood why they used `begin` here. Am I missing something?

spec/lib/zipkin_kafka_tracer_spec.rb
```diff
@@ -92,15 +88,15 @@ def initialize;end
...
       it 'sets the span name' do
         allow(id).to receive(:sampled?) { true }
-        allow(subject).to receive(:get_span_for_id) { span }
+        allow(tracer).to receive(:get_span_for_id) { span }
         expect(span).to receive(:name=).with(name)
-        subject.set_rpc_name(id, name)
+        tracer.set_rpc_name(id, name)
       end
     end
   end
```
jfeltesse-mdsol: thanks for getting rid of all those `subject` calls, they didn't make any sense but were out of the scope of my previous PRs so I left them untouched.

spec/support/test_app_config.ru
```diff
@@ -6,8 +6,8 @@ zipkin_tracer_config = {
   service_name: 'your service name here',
   service_port: 9410,
   sample_rate: 1,
-  scribe_max_buffer: 1,
-  scribe_server: '127.0.0.1:9410'
+  traces_buffer: 1,
+  log_api_host: '127.0.0.1:9410'
```
jfeltesse-mdsol: `log_api_host`? Did you mean `json_api_host`?

lib/zipkin-tracer/zipkin_kafka_tracer.rb
```diff
@@ -2,64 +2,30 @@
 require 'finagle-thrift/tracer'
 require 'hermann/producer'
 require 'hermann/discovery/zookeeper'
+require 'zipkin-tracer/zipkin_tracer_base'
 
 module Trace
-  class ZipkinKafkaTracer < Tracer
-    TRACER_CATEGORY     = "zipkin".freeze
+  # This class sends information to Zipkin throught Kafka.
```
jfeltesse-mdsol: `through`

lib/zipkin-tracer/zipkin_json_tracer.rb
```diff
@@ -19,55 +18,11 @@ def perform(json_api_host, spans)
...
-        Span.new("", id)
-      end
-    end
-
-    def reset
-      @count = 0
-      @spans = {}
-    end
-
+  # This class sends information to the Zipkin API,
```
jfeltesse-mdsol: that comma is odd

jfeltesse-mdsol:
not a huge fan of the name `zipkin_tracer_base` but can't come up with a better alt now. `zipkin_abstract_tracer` sounds Java-ish but would be a tad more meaningful IMHO.

lib/zipkin-tracer/zipkin_tracer_base.rb
jcarres-mdsol: rubocop is drunk, I do think there is any benefit of using fail instead of raise

jfeltesse-mdsol: you *do think*? :trollface: 

yeah that's debatable, their point is that "fail" should be used when provoking exceptions and "raise" should be used within the rescue blocks to re-raise.
I don't really care myself TBH.

jcarres-mdsol: Fail is a keyword. Untouchable.
"raise" is in fact Kernel.raise. It can be overwritten, if this gem is inside some projects which has let's say instrumentation and each raise does some kind of special notification or something, we would not be able to use that.

lib/zipkin-tracer/zipkin_tracer_base.rb
jcarres-mdsol: done

jcarres-mdsol:
@jfeltesse-mdsol  I've left it at zipkin_tracer_base , not best name ever but could not find anything significantly better

jfeltesse-mdsol:
LGTM, I'd press the button if I could...

PR #36, opened 2015-12-22
improve rack middleware specs

jcarres-mdsol:
The only change is that now we do not rescue Exception but where strictly necessary. 
Most of the PR is an improvement of the specs so we check for the type of annotations instead of just the number of times annotating. No functional changes so not upgrading the version now.

@jfeltesse-mdsol  @adriancole

### DIFF
--- a/lib/zipkin-tracer/rack/zipkin-tracer.rb
+++ b/lib/zipkin-tracer/rack/zipkin-tracer.rb
@@ -67,6 +67,14 @@ module ZipkinTracer
 
     private
 
+    def record(annotation)
+      ::Trace.record(annotation)
+    # Nothing wonky that the tracer does should stop us from using the app!!!
+    # Usually is better to rescue StandardError but the socket layer can launch Errno kind of exceptions
+    rescue Exception
+      # Ignore low level errors
+    end
+
     # Use the Domain environment variable to extract the service name, otherwise use the default config name
     def service_name(default_name)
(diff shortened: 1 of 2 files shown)


lib/zipkin-tracer/rack/zipkin-tracer.rb
```diff
@@ -124,9 +132,7 @@ def synchronize(&block)
       @lock.synchronize do
         yield
       end
-      # Nothing wonky that the tracer does should stop us from using the app!!!
-      # Usually is better to rescue StandardError but the socket layer can launch Errno kind of exceptions
-    rescue Exception => e
+    rescue StandardError => e
```
jfeltesse-mdsol: what about `rescue => e` ?

spec/lib/rack/zipkin-tracer_spec.rb
```diff
@@ -54,6 +74,12 @@ def mock_env(path = '/', params = {})
 
     context 'configured to use json' do
       subject { middleware(app, json_api_host: 'fake_json_api_host') }
+
+      let(:host_ip) { '127.0.0.1' }
+      before do
+        # A hack for our code to not be able to resolve localhost
+        allow(Socket).to receive(:getaddrinfo).and_raise StandardError
```
jfeltesse-mdsol: what about raising `SocketError` so it is closer to reality?

jfeltesse-mdsol:
:+1:

PR #38, opened 2015-12-25
thread safe always create IDs

jcarres-mdsol:
Kind of big,  Merry christmas!

@jfeltesse-mdsol  @adriancole

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 0.9.2
+Bugfix: Tracer is now Threadsafe
+Bugfix: Always create trace IDs even when the trace  will not be sent to zipkin (other parts of the app may use them).
+
 # 0.9.1
 Make Scribe actually optional by inspecting the conf first and requiring after.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
-# 0.9.2
+# 0.10.0
+Always create trace IDs even when the trace  will not be sent to zipkin (other parts of the app may use them).
 Bugfix: Tracer is now Threadsafe
-Bugfix: Always create trace IDs even when the trace  will not be sent to zipkin (other parts of the app may use them).
+Development improvement: Benchmark Rake task to help finding performance issues
 
 # 0.9.1
 Make Scribe actually optional by inspecting the conf first and requiring after.
--- a/README.md
+++ b/README.md
@@ -158,3 +158,29 @@ For example:
 # sample if request header specifies known device identifier
 lambda { |env| KNOWN_DEVICES.include?(env['HTTP_X_DEVICE_ID']) }
 ```
+
+## Development
+
+This project uses Rspec. Make sure your PRs contain proper tests.
+We have two rake task to help finding performance issues:
+```
+rake benchmark
+```
+Will run a benchmark testing all the different tracers and giving you
+their relative speed.
+
+```
+rake run_once
+```
+Will run the rack middleware, optionally the faraday middleware. Please
+modify the code to run the middleware you want to test.
+The best way to use this rake test is together with rbtrace.
+First run the task in background:
+```
+rake run_once &
+```
+Take note of the PID that displays in your terminal and run:
+```
+rbtrace -p PID -f
+```
+It will print out the methods used and the time each took.
--- a/Rakefile
+++ b/Rakefile
@@ -14,6 +14,13 @@
 
 require 'bundler/gem_tasks'
 require 'rspec/core/rake_task'
+require 'zipkin-tracer'
+require 'benchmark'
+require 'benchmark/ips'
+require 'faraday'
+require 'rack/mock'
+require 'rbtrace'
+require 'tempfile'
 
 def add_rspec_options(options=[])
   if RUBY_PLATFORM == 'java'
@@ -29,3 +36,111 @@ RSpec::Core::RakeTask.new(:spec) do |r|
 end
 
 task :default => :spec
+
+
+# Used to test a completely minimum middleware, no zipkin.
+class EmptyMiddleware
+  def initialize(app)
+    @app = app
+  end
+
+  def call(env)
+    if @app.respond_to?(:call)
+      status, headers, body = @app.call(env)
+    else
+      [200, {}, 'Hello world']
+    end
+  end
+end
+
+
+# Used to mock a middleware which uses Faraday.
+# Note that we are creating fake objects pretending to be Faraday for speed.
+class FaradayMiddleware
+  def initialize(_app)
+    response_env = { status: 200 }
+    @env = {
+        method: :post,
+        url: 'http://www.this_is_not_really_called.com',
+        body: 'majere',
+        request_headers: Faraday::Utils::Headers.new({}),
+      }
+      @app = lambda { |env| ResponseObject.new(@env, response_env) }
+  end
+
+  def call(_env)
+    middleware.call(@env)
+  end
+
+  private
+  def middleware
+    @middleware ||= ZipkinTracer::FaradayHandler.new(@app)
+  end
+
+  class ResponseObject
+    attr_reader :env
+
+    def initialize(env, response_env)
+      @env = env
+      @response_env = response_env
+    end
+
+    def on_complete
+      yield @response_env
+      self
+    end
+  end
+end
+
+desc "Runs a zipkin middleware once."
+task :run_once do
+  empty_app = EmptyMiddleware.new(nil)
+  app = FaradayMiddleware.new(nil)
+
+  logger = Logger.new(Tempfile.new('fakelog'))
+  null_configuration = { logger: logger, sample_rate: 1}
+  null_tracer_rack = ZipkinTracer::RackHandler.new(app, null_configuration)
+  env = Rack::MockRequest.env_for('/path', {})
+
+  sleep(10)
+  null_tracer_rack.call(env)
+end
+
+# This task is used to help development of ZipkinTracer.
+# It benchmark the relative performance of the different tracers.
+desc "Runs benchmarks for ZipkinTracer."
+task :benchmark do
+  logger = Logger.new(Tempfile.new('fakelog'))
+  fake_url = 'http://www.google.com' #resolve but unable to send I hope!
+
+  empty_app = EmptyMiddleware.new(nil)
+  app = FaradayMiddleware.new(nil)
+
+  null_configuration = { logger: logger, sample_rate: 1}
+  json_configuration = null_configuration.merge({ json_api_host: fake_url })
+  scribe_configuration = null_configuration.merge({ scribe_server: fake_url })
+
+  empty_rack = EmptyMiddleware.new(empty_app)
+  null_tracer_rack = ZipkinTracer::RackHandler.new(empty_app, null_configuration)
+  null_tracer_faraday_rack = ZipkinTracer::RackHandler.new(app, null_configuration)
+  json_tracer_rack = ZipkinTracer::RackHandler.new(empty_app, json_configuration)
+  json_tracer_faraday_rack = ZipkinTracer::RackHandler.new(app, json_configuration)
+  scribe_tracer_rack = ZipkinTracer::RackHandler.new(empty_app, scribe_configuration)
+  scribe_tracer_faraday_rack = ZipkinTracer::RackHandler.new(app, scribe_configuration)
+
+  env = Rack::MockRequest.env_for('/path', {})
+
+  Benchmark.ips do |bm|
+ #   bm.report("No rack middleware") { empty_rack.call(env) } # Uncomment if curious
+    bm.report("NullTracer") { null_tracer_rack.call(env) }
+    bm.report("NullTracer + Faraday") { null_tracer_faraday_rack.call(env) }
+    bm.report("JSONTracer") { json_tracer_rack.call(env) }
+    bm.report("JSONTracer + Faraday") { json_tracer_faraday_rack.call(env) }
+    bm.report("ScribeTracer") { scribe_tracer_rack.call(env) }
+    bm.report("ScribeTracer + Faraday") { scribe_tracer_faraday_rack.call(env) }
+    bm.compare!
+  end
+
+  puts "i/s means the number of times the middleware can be called per second"
+
+end
--- a/lib/zipkin-tracer/faraday/zipkin-tracer.rb
+++ b/lib/zipkin-tracer/faraday/zipkin-tracer.rb
@@ -21,28 +21,24 @@ module ZipkinTracer
     def call(env)
       # handle either a URI object (passed by Faraday v0.8.x in testing), or something string-izable
       url = env[:url].respond_to?(:host) ? env[:url] : URI.parse(env[:url].to_s)
-      local_endpoint = ::Trace.default_endpoint # The rack middleware set this up for us.
+      local_endpoint = Trace.default_endpoint # The rack middleware set this up for us.
       remote_endpoint = callee_endpoint(url, local_endpoint.ip_format) # The endpoint we are calling.
 
       response = nil
-      begin
-        trace_id = ::Trace.id
-        ::Trace.push(trace_id.next_id)
+      with_trace_id(Trace.id.next_id) do
         B3_HEADERS.each do |method, header|
-          env[:request_headers][header] = ::Trace.id.send(method).to_s
+          env[:request_headers][header] = Trace.id.send(method).to_s
         end
         # annotate with method (GET/POST/etc.) and uri path
-        ::Trace.set_rpc_name(env[:method].to_s.downcase)
-        record(::Trace::BinaryAnnotation.new('http.uri', url.path, 'STRING', local_endpoint))
-        record(::Trace::BinaryAnnotation.new('sa', '1', 'BOOL', remote_endpoint))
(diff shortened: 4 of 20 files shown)


lib/zipkin-tracer/version.rb
```diff
@@ -12,5 +12,5 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 module ZipkinTracer
-  VERSION = "0.9.1"
+  VERSION = "0.9.2"
```
adriancole: nit: seems pretty functional++ change to be a subminor

lib/zipkin-tracer/tracer_factory.rb
```diff
@@ -0,0 +1,40 @@
+
+module ZipkinTracer
```
adriancole: I like this refactor

lib/zipkin-tracer/version.rb
jcarres-mdsol: That's true, I'll do 0.10 + some other minor changes tomorrow

adriancole:
Nice work. Mind pinging back with concrete users of the trace id when not tracing? Also, can you paste an example benchmark run?

jcarres-mdsol:
@adriancole sure:
The benchmark:
```
Comparison:
           NullTracer:           57965.1 i/s
          JSONTracer:        11816.3 i/s - 4.91x slower
NullTracer + Faraday:     8759.4 i/s - 6.62x slower
JSONTracer + Faraday:  4426.6 i/s - 13.09x slower
        ScribeTracer:           3385.0 i/s - 17.12x slower
ScribeTracer + Faraday:  2036.9 i/s - 28.46x slower

i/s means the number of times the middleware can be called per second
```
About why create IDs even when not sending them to zipkin. This is the user-case:
We add the traceIDs to every log message in every app and send all logs to a central aggregator. Later we can search by traceID and retrieve all info related to any given trace. It is extremely useful to see that the error you got in one app relates to some other error happening in a different app. That connection is found thanks to always having traceIDs. 
So we always want to have the ID created even if we are not sampling this particular request.

adriancole:
thx, so in short, trace id is a log correlation id even if the trace isn't sampled in zipkin.

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 0.9.2
+Bugfix: Tracer is now Threadsafe
+Bugfix: Always create trace IDs even when the trace  will not be sent to zipkin (other parts of the app may use them).
+Development improvement: Benchmark Rake task to help finding performance
+issues
```
jfeltesse-mdsol: weird newline

README.md
```diff
@@ -158,3 +158,31 @@ For example:
...
+The best way to use this rake test is together with rbtrace.
+First run the task in background:
+```
+rake run_once &
+```
+Take note of the PID that displays in your terminal and run:
+```
+rbtrace -p PID -f
+```
+It will the methods used and their performance.
```
jfeltesse-mdsol: It will what?

Rakefile
```diff
@@ -29,3 +36,87 @@ RSpec::Core::RakeTask.new(:spec) do |r|
...
+  env = Rack::MockRequest.env_for('/path', {})
+
+  Benchmark.ips do |bm|
+ #   bm.report("No rack middleware") { empty_rack.call(env) } # Uncomment if curious
+    bm.report("NullTracer")    { null_tracer_rack.call(env) }
+    bm.report("NullTracer + Faraday") { null_tracer_faraday_rack.call(env) }
+    bm.report("JSONTracer")    { json_tracer_rack.call(env) }
+    bm.report("JSONTracer + Faraday") { json_tracer_faraday_rack.call(env) }
+    bm.report("ScribeTracer")    { scribe_tracer_rack.call(env) }
+    bm.report("ScribeTracer + Faraday") { scribe_tracer_faraday_rack.call(env) }
```
jfeltesse-mdsol: what's with the weird alignment of the blocks?

Rakefile
```diff
@@ -14,6 +14,13 @@
 
 require 'bundler/gem_tasks'
 require 'rspec/core/rake_task'
+require 'zipkin-tracer'
+require 'benchmark'
+require 'benchmark/ips'
+require 'faraday'
+require 'rack/mock'
+require 'rbtrace'
+require 'tempfile'
```
jfeltesse-mdsol: what about using a group in the Gemfile so you could just do `Bundle.require(:performance)` for instance?

README.md
```diff
@@ -158,3 +158,31 @@ For example:
...
+
+
```
jfeltesse-mdsol: see comment below about the group but if you go that way the commands would need to be `bundler exec` prefixed.

Rakefile
```diff
@@ -29,3 +36,87 @@ RSpec::Core::RakeTask.new(:spec) do |r|
...
+  end
+end
+
+class FaradayMiddleware
+  def initialize(app)
+    @app = app
+  end
+
+  def call(env)
+    conn = Faraday.new(url: 'http://www.google.com') do |faraday|
```
jfeltesse-mdsol: you're actually benchmarking against google.com? what about webmock or something? what if google is slow in your geographical region or something?

Rakefile
```diff
@@ -29,3 +36,87 @@ RSpec::Core::RakeTask.new(:spec) do |r|
...
+      faraday.use ZipkinTracer::FaradayHandler
+      # default Faraday stack
+      faraday.request :url_encoded
+      faraday.adapter Faraday.default_adapter
+    end
+   # conn.get('/')
```
jfeltesse-mdsol: that comment should be left out

Rakefile
```diff
@@ -29,3 +36,87 @@ RSpec::Core::RakeTask.new(:spec) do |r|
...
+  logger = Logger.new(Tempfile.new('fakelog'))
+  null_configuration = { logger: logger, sample_rate: 1}
+  null_tracer_rack = ZipkinTracer::RackHandler.new(app, null_configuration)
+  env = Rack::MockRequest.env_for('/path', {})
+
+  sleep(10)
+  null_tracer_rack.call(env)
+end
+
+# This task is used to help development of ZipkinTracer.
```
jfeltesse-mdsol: maybe that comment could be removed or turned into something more useful, like explain what kind of benchs it's doing etc.

Rakefile
```diff
@@ -29,3 +36,87 @@ RSpec::Core::RakeTask.new(:spec) do |r|
...
+  null_tracer_faraday_rack = ZipkinTracer::RackHandler.new(app, null_configuration)
+  json_tracer_rack = ZipkinTracer::RackHandler.new(empty_app, json_configuration)
+  json_tracer_faraday_rack = ZipkinTracer::RackHandler.new(app, json_configuration)
+  scribe_tracer_rack = ZipkinTracer::RackHandler.new(empty_app, scribe_configuration)
+  scribe_tracer_faraday_rack = ZipkinTracer::RackHandler.new(app, scribe_configuration)
+
+  env = Rack::MockRequest.env_for('/path', {})
+
+  Benchmark.ips do |bm|
+ #   bm.report("No rack middleware") { empty_rack.call(env) } # Uncomment if curious
```
jfeltesse-mdsol: remove

lib/zipkin-tracer/faraday/zipkin-tracer.rb
```diff
@@ -19,30 +19,28 @@ def initialize(app, service_name = nil)
...
       url = env[:url].respond_to?(:host) ? env[:url] : URI.parse(env[:url].to_s)
-      local_endpoint = ::Trace.default_endpoint # The rack middleware set this up for us.
+      local_endpoint = Trace.default_endpoint # The rack middleware set this up for us.
       remote_endpoint = callee_endpoint(url, local_endpoint.ip_format) # The endpoint we are calling.
 
       response = nil
-      begin
-        trace_id = ::Trace.id
-        ::Trace.push(trace_id.next_id)
+      @trace_id = Trace.id.next_id # Use a var instead of Trace.id for speed
```
jfeltesse-mdsol: IMHO you don't need to justify yourself here :trollface:

jfeltesse-mdsol: and why do you use an instance var, as opposed to a local var (e.g. without the `@` prefix)?

lib/zipkin-tracer/rack/zipkin-tracer.rb
```diff
@@ -25,59 +27,33 @@ class RackHandler
...
-        nil, # auto detect hostname
-        config.service_port,
-        service_name(config.service_name),
-        ip_format
-      )
-      ::Trace.sample_rate = config.sample_rate
-
-      @config = config
+      @config = Config.new(app, config).freeze
+      @tracer = TracerFactory.new.tracer(@config)
```
jfeltesse-mdsol: how about having `attr_reader` for those 2 so you don't manipulate the instance var directly?

lib/zipkin-tracer/rack/zipkin-tracer.rb
```diff
@@ -25,59 +27,33 @@ class RackHandler
...
 
     def call(env)
-      # skip certain requests
-      return @app.call(env) if filtered?(env) || !routable_request?(env)
-
-      whitelisted = force_sample?(env)
-      id = get_or_create_trace_id(env, whitelisted) # note that this depends on the sample rate being set
-      tracing_filter(id, env, whitelisted) { @app.call(env) }
+      zipkin_env = ZipkinEnv.new(env, @config)
+      @trace_id = zipkin_env.trace_id
```
jfeltesse-mdsol: likewise, you don't need an instance var here, do you?

jfeltesse-mdsol: mmm I see it's being implicitly used by `trace!` and `trace_request_information`, wondering if the trace_id var shouldn't be passed as an arg. your call.

lib/zipkin-tracer/rack/zipkin-tracer.rb
```diff
@@ -136,19 +97,55 @@ def synchronize(&block)
...
+      private
+
+      def current_trace_sampled?
+        rand < @config.sample_rate
+      end
 
-      trace_parameters += env.values_at(*B3_OPT_HEADERS) # always check flags
-      trace_parameters[4] = (trace_parameters[4] || default_flags).to_i
+      def should_trace?(parent_trace_sampled)
+        if parent_trace_sampled  #A service upstream decided this goes in all the way
```
jfeltesse-mdsol: [nits] no space after "#" and capital "A"

lib/zipkin-tracer/tracer_factory.rb
```diff
@@ -0,0 +1,40 @@
...
+          Trace::ZipkinTracer.new(CarelessScribe.new(config.scribe_server), config.scribe_max_buffer)
+        when :kafka
+          require 'zipkin-tracer/zipkin_kafka_tracer'
+          Trace::ZipkinKafkaTracer.new(zookeepers: config.zookeeper)
+        else
+          Trace::NullTracer.new
+      end
+      Trace.tracer = tracer
+
+      # TODO: move this to the TracerBase and kill scribe tracer
```
jfeltesse-mdsol: what's the ref about killing scribe about? IIRC kafka uses :i32 too.

lib/zipkin-tracer/zipkin_json_tracer.rb
```diff
@@ -21,8 +22,15 @@ module Trace
...
+
+    def initialize(options)
+      SuckerPunch.logger = options[:logger]
+      @json_api_host = options[:json_api_host]
+      super(options)
+    end
+
     def flush!
-      AsyncJsonApiClient.new.async.perform(@options[:json_api_host], @spans.values.dup)
+      AsyncJsonApiClient.new.async.perform(@json_api_host, spans.values.dup)
```
jfeltesse-mdsol: json_api_host should be accessed through an accessor IMHO

lib/zipkin-tracer/zipkin_kafka_tracer.rb
```diff
@@ -19,14 +19,15 @@ def initialize(options = {})
...
     def flush!
-      topic = opts[:topic] || DEFAULT_KAFKA_TOPIC
-      messages = @spans.values.map do |span|
+      messages = spans.values.map do |span|
         buf = ''
         trans = Thrift::MemoryBufferTransport.new(buf)
         oprot = Thrift::BinaryProtocol.new(trans)
         span.to_thrift.write(oprot)
-        @producer.push(buf, :topic => topic).value!
+        @producer.push(buf, :topic => @topic).value!
```
jfeltesse-mdsol: time to use the modern hash syntax IMO

spec/lib/rack/zipkin-tracer_spec.rb
```diff
@@ -3,7 +3,8 @@
 
 describe ZipkinTracer::RackHandler do
   def middleware(app, config={})
-    described_class.new(app, { logger: Logger.new(nil) }.merge(config))
+    configuration = { logger: logger, sample_rate: 1}.merge(config)
```
jfeltesse-mdsol: [nits] lacks space before closing and you could use `merge!` here.

Rakefile
jcarres-mdsol: I think it is useful for someone trying to know the performance hit in general. Comparing with a rack middleware doing nothign allows that. I do not leave it uncommented because usually everything looks so slow in comparison and it is difficult to compare zipkin implementations

lib/zipkin-tracer/faraday/zipkin-tracer.rb
jcarres-mdsol: The record method was using it. I can get rid of that method and the ivar also thought

spec/lib/tracer_factory_spec.rb
```diff
@@ -0,0 +1,88 @@
+require 'rack/mock'
+require 'spec_helper'
+
+describe ZipkinTracer::TracerFactory do
+  def middleware(app, config={})
+    configuration = { logger: logger, sample_rate: 1}.merge(config)
```
jfeltesse-mdsol: same comment as above

lib/zipkin-tracer/rack/zipkin-tracer.rb
jcarres-mdsol: I'll do, not really like ivars

spec/lib/tracer_factory_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+        it 'creates a zipkin kafka tracer' do
+          allow(::Trace::ZipkinKafkaTracer).to receive(:new) { tracer }
+          expect(::Trace).to receive(:tracer=).with(tracer)
+          described_class.new.tracer(config)
+        end
+      end
+    end
+
+    context 'configured to use json' do
+      let(:config) { configuration({json_api_host: 'fake_json_api_host'}) }
```
jfeltesse-mdsol: useless '{' '}'

lib/zipkin-tracer/zipkin_json_tracer.rb
jcarres-mdsol: IMHO attr accessors become the public API . Specially being a gem we do not know who may be accesing it for what reasons. If possible I want to keep things private.

spec/lib/zipkin_json_tracer_spec.rb
```diff
@@ -12,16 +12,10 @@
 
   let(:json_api_host) { 'http://json.example.com' }
   let(:traces_buffer) { 1 }
-  let(:tracer) { described_class.new(json_api_host: json_api_host, traces_buffer: traces_buffer) }
+  let(:default_options) { {json_api_host: json_api_host, traces_buffer: traces_buffer} }
```
jfeltesse-mdsol: spaces within the braces

spec/lib/zipkin_json_tracer_spec.rb
```diff
@@ -82,17 +76,17 @@
     end
   end
 
+  describe '#initialize' do
+    let(:logger) { nil }
+    it 'sets the SuckerPunch logger' do
+      expect(SuckerPunch).to receive(:logger=).with(logger)
+      described_class.new(default_options.merge({logger: logger}))
```
jfeltesse-mdsol: useless braces

zipkin-tracer.gemspec
```diff
@@ -42,8 +42,13 @@ Gem::Specification.new do |s|
...
   s.add_development_dependency 'timecop', '~> 0.8'
   s.add_development_dependency 'webmock', '~> 1.22'
+  s.add_development_dependency 'mutant-rspec'
+  s.add_development_dependency 'byebug'
 
   # manually install the gem depending on the adapter you want to use
   s.add_development_dependency 'hermann', '~> 0.25'
   s.add_development_dependency 'scribe', '~> 0.2.4'
+  s.add_development_dependency 'benchmark-ips'
+  s.add_development_dependency 'rbtrace'
```
jfeltesse-mdsol: those gems should be in a group in the Gemfile (see previous comments) that you'd only need if you want to run the benchmarks.
at the very least these shouldn't be in that "block" because then the comment at line 48 doesn't make sense anymore.

lib/zipkin-tracer/rack/zipkin-tracer.rb
jcarres-mdsol: I do not like attr_reader on anything that is not a public interface

PR #39, opened 2015-12-28
Update the travis configuration

jfeltesse-mdsol:
Remove the bundler hack from rev 3e92f619 now that the regression has been fixed


jfeltesse-mdsol:
@jcarres-mdsol @adriancole

PR #40, opened 2016-01-07
only create binaryannotations and annotations when we are sampling

jcarres-mdsol:
Avoid doing any tracing related business if we are not sampling this request.
if the sampling rate is 1% we are creating annotation objects and doing some processing 99% of the time to just throw it away afterwards. This PRs aims at solving that and hopefully make tracing lighter to its users.

@jfeltesse-mdsol  @adriancole

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -57,7 +57,7 @@ end
 # Used to mock a middleware which uses Faraday.
 # Note that we are creating fake objects pretending to be Faraday for speed.
 class FaradayMiddleware
-  def initialize(_app)
+  def initialize
     response_env = { status: 200 }
     @env = {
         method: :post,
@@ -66,16 +66,14 @@ class FaradayMiddleware
         request_headers: Faraday::Utils::Headers.new({}),
       }
       @app = lambda { |env| ResponseObject.new(@env, response_env) }
+      @middleware = ZipkinTracer::FaradayHandler.new(@app)
   end
 
   def call(_env)
-    middleware.call(@env)
+    @middleware.call(@env)
   end
 
   private
-  def middleware
-    @middleware ||= ZipkinTracer::FaradayHandler.new(@app)
-  end
 
   class ResponseObject
     attr_reader :env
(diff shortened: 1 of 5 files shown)


lib/zipkin-tracer/faraday/zipkin-tracer.rb
```diff
@@ -20,32 +20,40 @@ def initialize(app, service_name = nil)
...
-        @tracer.set_rpc_name(trace_id, env[:method].to_s.downcase)
-        @tracer.record(trace_id, Trace::BinaryAnnotation.new('http.uri', url.path, 'STRING', local_endpoint))
-        @tracer.record(trace_id, Trace::BinaryAnnotation.new('sa', '1', 'BOOL', remote_endpoint))
-        @tracer.record(trace_id, Trace::Annotation.new(Trace::Annotation::CLIENT_SEND, local_endpoint))
-        response = @app.call(env).on_complete do |renv|
-          # record HTTP status code on response
-          @tracer.record(trace_id, Trace::BinaryAnnotation.new('http.status', renv[:status].to_s, 'STRING', local_endpoint))
+        if trace_id.sampled?
+          trace!(env, trace_id)
+        else
```
adriancole: I'd propagate the sampled decision downstream if possible. `X-B3-Sampled` header 0 or 1, so that downstream calls aren't mistaken for new root traces. (assuming this isn't done already)

spec/lib/faraday/zipkin-tracer_spec.rb
```diff
@@ -122,6 +124,48 @@ def expect_tracing
...
+        result = nil
+        ::Trace.push(trace_id) do
+          result = process('', url).env
+        end
+
+        expect(result[:request_headers]['X-B3-TraceId']).to eq('0000000000000001')
+        expect(result[:request_headers]['X-B3-ParentSpanId']).to eq('0000000000000003')
+        expect(result[:request_headers]['X-B3-SpanId']).not_to eq('0000000000000003')
+        expect(result[:request_headers]['X-B3-SpanId']).to match(HEX_REGEX)
+        expect(result[:request_headers]['X-B3-Sampled']).to eq('false')
```
adriancole: oh here it is

adriancole:
LGTM conceptually

Rakefile
```diff
@@ -114,24 +112,28 @@ task :benchmark do
...
   empty_app = EmptyMiddleware.new(nil)
-  app = FaradayMiddleware.new(nil)
 
   null_configuration = { logger: logger, sample_rate: 1}
   json_configuration = null_configuration.merge({ json_api_host: fake_url })
   scribe_configuration = null_configuration.merge({ scribe_server: fake_url })
 
+  # We create a different faraday middleware per rack middleware below because
+  # both middlewares share the same tracer. So they need to be created in pairs.
   empty_rack = EmptyMiddleware.new(empty_app)
```
jfeltesse-mdsol: me thinks that for consistency that class should also remove the `app` argument and do what the faraday class does now.

lib/zipkin-tracer/faraday/zipkin-tracer.rb
```diff
@@ -20,32 +20,40 @@ def initialize(app, service_name = nil)
...
-      response
     end
 
     private
 
+    def trace!(env, trace_id)
+      response = nil
+      # handle either a URI object (passed by Faraday v0.8.x in testing), or something string-izable
+      url = env[:url].respond_to?(:host) ? env[:url] : URI.parse(env[:url].to_s)
+      local_endpoint = Trace.default_endpoint # The rack middleware set this up for us.
```
jfeltesse-mdsol: `sets`

lib/zipkin-tracer/faraday/zipkin-tracer.rb
```diff
@@ -20,32 +20,40 @@ def initialize(app, service_name = nil)
...
+      remote_endpoint = callee_endpoint(url, local_endpoint.ip_format) # The endpoint we are calling.
+      # annotate with method (GET/POST/etc.) and uri path
+      @tracer.set_rpc_name(trace_id, env[:method].to_s.downcase)
+      @tracer.record(trace_id, Trace::BinaryAnnotation.new('http.uri', url.path, 'STRING', local_endpoint))
```
jfeltesse-mdsol: I think those STRING and BOOL strings should be constants.
That would prevent typos and consume a tad less memory. IIRC rack does this for the strings often used, like "Content-Type" and such.

jfeltesse-mdsol: Applies for the `http.uri` and other `sa` too. 
Especially in the `sa` case it would allow us to have an easier to understand constant name than this acronym.

jfeltesse-mdsol:
:ok_woman:

Rakefile
jcarres-mdsol: I'm doing two different things there depending on if we pass an app or nil, maybe should be refactored on two classes

jcarres-mdsol:
added the constants and merging

PR #41, opened 2016-01-14
Add faraday dependency

jfeltesse-mdsol:
The base tracer class requires it, raising a LoadError if it's not part of the app's bundle.

Reproducing is easy, simply make a brand new app without including faraday in your Gemfile and try to run it, it will raise
```
lib/zipkin-tracer/zipkin_tracer_base.rb:1:in `require': cannot load such file -- faraday (LoadError)
```

@jcarres-mdsol @adriancole

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 0.10.2
+Add faraday as a dependency
+
 # 0.10.1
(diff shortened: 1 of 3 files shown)


adriancole:
sounds reasonable

jcarres-mdsol:
cool, thanks!

PR #42, opened 2016-01-14
avoid depending on finagle-thrift and thrif if possible

jcarres-mdsol:
Turns out we were using a file which depends on thrift. Taking that out would make an app blow up at the missing dependency.
Making requires more specific so we only require thrift when using scribe, not if we use the json api

@jfeltesse-mdsol  @adriancole


jcarres-mdsol:
I'm going to merge this, kind of an urgency. we can always make a 0.10.4 if something was wrong with these changes.

PR #43, opened 2016-01-22
Add Local Tracing

ykitamura-mdsol:
Implement local trace for Zipkin.
I use "record_tag" for the method name to record binary annotation.

Specs and documents are also updated.
Please review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 0.10.4
+Add local tracing
+
 # 0.10.3
 Avoid requiring finagle-thrift when possible to avoid a hard dependency on Thrift
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
-# 0.11.1
-Add local tracing
+# 0.12.0
+Add local tracing, fix flushing, add timestamp and duration to span
 
 # 0.11.0
 Use local spans instead of thread-safe variables to improve performance
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 0.11.1
+Add local tracing
+
 # 0.11.0
 Use local spans instead of thread-safe variables to improve performance
 Add new "with_new_span" method to the tracer api to allow creating
--- a/README.md
+++ b/README.md
@@ -58,6 +58,41 @@ Note that supplying the service name for the destination service is optional;
 the tracing will default to a service name derived from the first section of the destination URL (e.g. 'service.example.com' => 'service').
 
 
+### Local tracing
+
+`ZipkinTracer::TraceClient` provides API to record local traces in your application.
+
+Both examples below have the same result.
+
+Example 1:
+```ruby
+ZipkinTracer::TraceClient.trace do |ztc|
+  ztc.record_local_component 'Local Trace'
+  ztc.record_tag 'key', 'sample'
+  # target process
+end
+```
+
+Example 2:
+```ruby
+ztc = ZipkinTracer::TraceClient
+ztc.record 'Start: '
+ztc.record_local_component 'Local Trace'
+ztc.record_tag 'key', 'sample'
+# target process
+ztc.record 'End: '
+```
+
+`.trace` method accepts one optional parameter to show in annotation keys.
+```ruby
+# with parameter
+ZipkinTracer::TraceClient.trace('services') do   # => 'Start: services', 'End: services'
+
+# without parameter
+ZipkinTracer::TraceClient.trace do   # => 'Start: ', 'End: '
+```
+
+
 ## Tracers
 
 ### JSON
--- a/README.md
+++ b/README.md
@@ -60,36 +60,23 @@ the tracing will default to a service name derived from the first section of the
 
 ### Local tracing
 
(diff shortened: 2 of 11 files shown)


lib/zipkin-tracer/trace_client.rb
```diff
@@ -0,0 +1,26 @@
...
+
+    def self.record_tag(key, value)
+      Trace.record(Trace::BinaryAnnotation.new(key, value, STRING_TYPE, Trace.default_endpoint))
+    end
+
+    def self.record_local_component(value)
+      record_tag(LOCAL_COMPONENT, value)
+    end
+
+    def self.trace(key = nil)
```
jcarres-mdsol: I think this is a very good idea and I like it very much but then you should be creating new spans instead.
It will kind of give you the start/end in a more meaninful way and it will also show beautiful in the zipkin server.

You will need to check how to create new spans, maybe it's not easy as I think that's inside thrift code but very worth it

lib/zipkin-tracer/trace_client.rb
```diff
@@ -0,0 +1,26 @@
+module ZipkinTracer
+  class TraceClient
+    LOCAL_COMPONENT = 'lc'.freeze
+    STRING_TYPE = 'STRING'.freeze
```
jcarres-mdsol: nice

lib/zipkin-tracer/trace_client.rb
```diff
@@ -0,0 +1,26 @@
...
+
+    def self.record(key)
+      Trace.record(Trace::Annotation.new(key, Trace.default_endpoint))
+    end
+
+    def self.record_tag(key, value)
+      Trace.record(Trace::BinaryAnnotation.new(key, value, STRING_TYPE, Trace.default_endpoint))
+    end
+
+    def self.record_local_component(value)
```
jcarres-mdsol: cool, easy to use

spec/lib/trace_client_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    describe '.record_tag' do
+      it 'records a binary annotation' do
+        expect(Trace).to receive(:record).with(instance_of(Trace::BinaryAnnotation))
+        subject.record_tag(key, value)
+      end
+    end
+
+    describe '.record_local_component' do
+      it 'records a binary annotation ' do
+        expect(Trace).to receive(:record).with(instance_of(Trace::BinaryAnnotation))
```
jcarres-mdsol: maybe you can also check that the BinaryAnnotation got the proper parameters. I believe the specs of the faraday_middleware is doing that, you can copy it from there.

README.md
```diff
@@ -58,6 +58,41 @@ Note that supplying the service name for the destination service is optional;
...
+  ztc.record_local_component 'Local Trace'
+  ztc.record_tag 'key', 'sample'
+  # target process
+end
+```
+
+Example 2:
+```ruby
+ztc = ZipkinTracer::TraceClient
+ztc.record 'Start: '
```
jcarres-mdsol: As per my comment below, I think it would be better to create a new span instead of annotating start, end places in the current span

jcarres-mdsol:
Only an implementation problem, it's better to have a new span than start/end annotations.
Code-wise very good

README.md
```diff
@@ -58,6 +58,41 @@ Note that supplying the service name for the destination service is optional;
 the tracing will default to a service name derived from the first section of the destination URL (e.g. 'service.example.com' => 'service').
 
 
+### Local tracing
+
+`ZipkinTracer::TraceClient` provides API to record local traces in your application.
```
ssteeg-mdsol: Maybe a one-liner about how a local trace is different from a normal trace would be helpful (for those new to the project)

jfeltesse-mdsol: also, me thinks the copy could be improved, it doesn't read very natural.

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 0.11.1
```
jcarres-mdsol: 0.12

lib/zipkin-tracer/zipkin_tracer_base.rb
```diff
@@ -17,14 +17,17 @@ def initialize(options={})
 
     def with_new_span(trace_id, name)
       span = start_span(trace_id, name)
+      time_now = Time.now
+      span.timestamp = (time_now.to_f * 1_000_000).to_i # to microseconds
       result = yield span
+      span.duration = ((Time.now - time_now) * 1_000_000).to_i # to microseconds
```
jfeltesse-mdsol: - `time_now` is not a very good name, it should be for instance `start_time` or something.
- if we are to do this in other tracers it might be good to have some shared method so that everyone uses the same logic. at the minimum there should be some "to_microseconds" method. if possible a method to wrap the whole thing, doing the time math around the block would be cool.

spec/lib/trace_client_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+
+      subject.local_component_span(lc_value) do |ztc|
+        ztc.record(value)
+      end
+    end
+  end
+end
+
+
+
```
jfeltesse-mdsol: rubocop won't be happy with those trailing lines

PR #44, opened 2016-01-25
use spans instead of tracers

jcarres-mdsol:
Using spans everywhere, I think logically makes more sense and avoiding using Thread.current gives us a good performance boost.

We went from 100ns in the JSON rack middleware down to 28ns.

@jfeltesse-mdsol  @adriancole @ykitamura-mdsol 
tell me what do you think of this approach

### DIFF
--- a/lib/zipkin-tracer/tracer_factory.rb
+++ b/lib/zipkin-tracer/tracer_factory.rb
@@ -1,4 +1,3 @@
-
 module ZipkinTracer
   class TracerFactory
     def tracer(config)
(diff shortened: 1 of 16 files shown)


lib/zipkin-tracer/tracer_factory.rb
```diff
@@ -1,3 +1,27 @@
...
+
+    def flush!
+      @scribe.batch do
+        messages = spans.values.map do |span|
+          buf = ''
+          trans = Thrift::MemoryBufferTransport.new(buf)
+          oprot = Thrift::BinaryProtocol.new(trans)
+          span.to_thrift.write(oprot)
+          binary = Base64.encode64(buf).gsub("\n", "")
+          @scribe.log(binary, TRACER_CATEGORY)
```
adriancole: I think you want to make one call to log here.. it appears like this block makes a call for each span. scribe can accept a list of log entries in a single network call.

lib/zipkin-tracer/zipkin_scribe_tracer.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    def flush!
+      @scribe.batch do
+        messages = spans.values.map do |span|
+          buf = ''
+          trans = Thrift::MemoryBufferTransport.new(buf)
+          oprot = Thrift::BinaryProtocol.new(trans)
+          span.to_thrift.write(oprot)
+          binary = Base64.encode64(buf).gsub("\n", "")
+          @scribe.log(binary, TRACER_CATEGORY)
```
adriancole: same comment

adriancole:
overall comment that adding to a span is more explicit, and more like work going on in opentracing. Sound like a positive move.

ykitamura-mdsol:
LGTM!

Can I use the "with_new_span" method from local tracer like this?

```ruby
Trace.tracer.with_new_span(Trace.id, name) do |span|
```

PR #45, opened 2016-02-01
move code to its own classes

jcarres-mdsol:
No functional changes. I was playing with this gem https://github.com/whitesmith/rubycritic and it told me this code needed to be moved to a better class and some comments should be added to classes. I think it does make the code cleaner so I followed the recommendations.

@jfeltesse-mdsol

### DIFF
--- a/spec/lib/application_spec.rb
+++ b/spec/lib/application_spec.rb
@@ -1,0 +1,39 @@
+require 'spec_helper'
+module ZipkinTracer
+  RSpec.describe Application do
+    describe '.routable_request?' do
+      it 'returns rails routable information if available' do
+        stub_const("Rails", Class.new)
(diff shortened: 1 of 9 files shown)


spec/lib/application_spec.rb
```diff
@@ -0,0 +1,39 @@
+require 'spec_helper'
+module ZipkinTracer
+  RSpec.describe Application do
+    describe '.routable_request?' do
+      it 'returns rails routable information if available' do
+        stub_const("Rails", Class.new)
+        allow(Rails).to receive_message_chain("application.routes.recognize_path") { nil }
+        expect(Application.routable_request?("path")).to eq(true)
```
jfeltesse-mdsol: [nits] what about `(...).to be true` ?

jfeltesse-mdsol:
:+1:

PR #48, opened 2016-02-04
Make local tracing method returns the result of block

ykitamura-mdsol:
Modify local tracing method to return the result of block.

Now you can just wrap process even if you want to use the result.

```ruby
    def self.all
      all_items = Resource.items
      add_information(all_items)
      all_items
    end
```
:arrow_down:

```ruby
    def self.all
      ZipkinTracer::TraceClient.local_component_span("Services Data") do
        all_items = Resource.items
        add_information(all_items)
        all_items
      end
    end
```

Please review.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,10 +69,11 @@ When `local_component_span` method is called, it creates a new span and a local 
 
 Example:
 ```ruby
-ZipkinTracer::TraceClient.local_component_span('Local Trace') do |ztc|
-  ztc.record 'New Annotation'
-  ztc.record_tag 'key', 'sample'
-  # target process
+ZipkinTracer::TraceClient.local_component_span('DB process') do |ztc|
+  ztc.record 'Create users'
+  ztc.record_tag 'number', '1000'
+
+  # create 1000 users
 end
 ```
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -69,10 +69,13 @@ When `local_component_span` method is called, it creates a new span and a local
...
-  ztc.record 'New Annotation'
-  ztc.record_tag 'key', 'sample'
-  # target process
+ZipkinTracer::TraceClient.local_component_span('DB process') do |ztc|
+  ztc.record 'Create users'
+  ztc.record_tag 'number', '1000'
+
+  ActiveRecord::Base.transaction do
+    1000.times { User.create(options) }
+  end
```
jfeltesse-mdsol: I think this example serves no purpose but to make things more confusing and there are people out there not using ActiveRecord.
IMO it should be simply something like 
```ruby
# create 1000 users
```

lib/zipkin-tracer/trace_client.rb
```diff
@@ -5,19 +5,22 @@ class TraceClient
 
     def self.local_component_span(local_component_value, &block)
       if block_given?
```
jfeltesse-mdsol: why is this `if` needed? the block is in the args, if someone calls it without providing it an exception would be raised:
```ruby
def a(b, &block)
  puts b
  yield
end

a("this is b") { puts "from block" }
# => this is b
# => from block

a("this is b")
#=> this is b
#=> LocalJumpError: no block given (yield)
```

jcarres-mdsol: agree, this method needs a block always

PR #49, opened 2016-02-04
remove support for traces buffering

jcarres-mdsol:
Deleting support for buffers because we do not do it properly and we want to flush at the end of a request anyways. If ever the support comes back we would need to rethink it anyways.

Also moved some specs around, figured out they were in the wrong place.

@jfeltesse-mdsol  @ykitamura-mdsol  @adriancole

### DIFF
--- a/lib/zipkin-tracer/trace.rb
+++ b/lib/zipkin-tracer/trace.rb
@@ -18,19 +18,28 @@ module Trace
   # A span may contain many annotations
   # This class is defined in finagle-thrift. We are adding extra methods here
   class Span
-    attr_reader :size
-    attr_accessor :timestamp, :duration
+    def initialize(name, span_id)
+      @name = name
+      @span_id = span_id
+      @annotations = []
+      @binary_annotations = []
+      @debug = span_id.debug?
+      @timestamp = to_microseconds(Time.now)
+      @duration = UNKNOWN_DURATION
+    end
+
+    def close_span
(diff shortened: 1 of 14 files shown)


lib/zipkin-tracer/zipkin_tracer_base.rb
```diff
@@ -11,23 +11,18 @@ class ZipkinTracerBase < Tracer
...
-      may_flush(span)
+      end_span(span)
       result
     end
 
-    def may_flush(span)
-      size = spans.values.map(&:size).map(&:to_i).inject(:+) || 0
-      if size >= @traces_buffer || span.annotations.any?{ |ann| ann.value == Annotation::SERVER_SEND }
+    def end_span(span)
+      if span.annotations.any?{ |ann| ann.value == Annotation::SERVER_SEND }
```
ykitamura-mdsol: You might need to call `span.close_span` here

lib/zipkin-tracer/trace.rb
```diff
@@ -41,11 +50,18 @@ def to_h
...
-        timestamp: timestamp,
-        duration: duration,
+        timestamp: @timestamp,
+        duration: @duration,
         debug: @debug
       }
     end
+
+    private
+    UNKNOWN_DURATION = 0 # mark duration was not set
```
jfeltesse-mdsol: [nits] open a line before maybe?

PR #51, opened 2016-02-05
storing in an array instead of a hash

jcarres-mdsol:
Now that we never retrieve individual spans but all of them at the end of the request, we can use an array directly instead of a hash and then extract the values as array. 

@jfeltesse-mdsol  @ykitamura-mdsol


jfeltesse-mdsol:
LGTM

PR #53, opened 2016-02-05
Use simplecov

jfeltesse-mdsol:
As we're at it, remove dead spec code that was revealed by the coverage report

@jcarres-mdsol @ykitamura-mdsol @ssteeg-mdsol

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,3 +1,8 @@
+if ENV['COV']
+  require 'simplecov'
+  SimpleCov.start
+end
+
 require 'zipkin-tracer'
 require 'timecop'
 require 'webmock/rspec'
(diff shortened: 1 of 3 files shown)


spec/spec_helper.rb
```diff
@@ -1,3 +1,6 @@
+require 'simplecov'
+SimpleCov.start
```
jcarres-mdsol: Can you protect Simplecov.start with  `if ENV['COV']` or `unless ENV['NO_COV']` or similar?
I'm usually using mutant with this gem, mutant runs hundreds of times the specs and it is a PITA to run coverage each time

jfeltesse-mdsol: done

jcarres-mdsol:
Cool, thanks!

PR #54, opened 2016-02-05
Move record methods in TraceClient to Span

ykitamura-mdsol:
Refactoring Span class, moving record methods in TraceClient class to Span. Update other classes to use the new methods.

Also, relocate BinaryAnnotation's constant variables to appropriate position. BinaryAnnotation types are defined here (http://www.rubydoc.info/gems/finagle-thrift/1.4.1/Trace/BinaryAnnotation/Type).

Please review  

### DIFF
--- a/lib/zipkin-tracer/faraday/zipkin-tracer.rb
+++ b/lib/zipkin-tracer/faraday/zipkin-tracer.rb
@@ -35,12 +35,7 @@ module ZipkinTracer
     end
 
     private
-    SERVER_ADDRESS = 'sa'.freeze
     SERVER_ADDRESS_SPECIAL_VALUE = '1'.freeze
-    STRING_TYPE = 'STRING'.freeze
-    BOOLEAN_TYPE = 'BOOL'.freeze
-    URI_KEY = 'http.uri'.freeze
-    STATUS_KEY = 'http.status'.freeze
 
     def trace!(env, trace_id)
       response = nil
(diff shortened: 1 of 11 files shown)


lib/zipkin-tracer/trace.rb
```diff
@@ -56,10 +46,32 @@ def to_h
...
+
+    def record_local_component(value)
+      record_tag(BinaryAnnotation::LOCAL_COMPONENT, value)
+    end
+
     private
 
     UNKNOWN_DURATION = 0 # mark duration was not set
 
+    def stack(annotation)
```
jcarres-mdsol: maybe `add` is better name, stack makes you think of a stack which is not

lib/zipkin-tracer/trace.rb
```diff
@@ -56,10 +46,32 @@ def to_h
...
+      case annotation
+      when BinaryAnnotation
+        binary_annotations << annotation
+      when Annotation
+        annotations << annotation
+      end
+    end
```
jfeltesse-mdsol: IMHO now the methods above are clearly separate I don't think we need to have this abstraction, we can add to the respective arrays directly in the methods, e.g. `annotations << Trace::Annotation.new(value, endpoint)` in `record` and `binary_annotations << Trace::BinaryAnnotation.new(key, value, type, endpoint)` in `record_tag`

lib/zipkin-tracer/trace.rb
jfeltesse-mdsol: @jcarres-mdsol see my comment below, what do you think? remove that method altogether I say.

jcarres-mdsol: you are right, better delete

lib/zipkin-tracer/faraday/zipkin-tracer.rb
```diff
@@ -50,14 +45,14 @@ def trace!(env, trace_id)
       remote_endpoint = Trace::Endpoint.remote_endpoint(url, @service_name, local_endpoint.ip_format) # The endpoint we are calling.
       @tracer.with_new_span(trace_id, env[:method].to_s.downcase) do |span|
         # annotate with method (GET/POST/etc.) and uri path
-        span.record(Trace::BinaryAnnotation.new(URI_KEY, url.path, STRING_TYPE, local_endpoint))
-        span.record(Trace::BinaryAnnotation.new(SERVER_ADDRESS, SERVER_ADDRESS_SPECIAL_VALUE, BOOLEAN_TYPE, remote_endpoint))
-        span.record(Trace::Annotation.new(Trace::Annotation::CLIENT_SEND, local_endpoint))
+        span.record_tag(Trace::BinaryAnnotation::URI, url.path, Trace::BinaryAnnotation::Type::STRING, local_endpoint)
```
jcarres-mdsol: URI not really belongs to BinaryAnnotation, same with the type string, should go to some other namespace

lib/zipkin-tracer/trace.rb
ykitamura-mdsol: OK!

lib/zipkin-tracer/trace_client.rb
```diff
@@ -1,8 +1,10 @@
 module ZipkinTracer
-  class TraceClient
-    LOCAL_COMPONENT = 'lc'.freeze
-    STRING_TYPE = 'STRING'.freeze
+  class NullSpan
+    def method_missing(name, *args)
```
jcarres-mdsol: that's cool, makes it simple

lib/zipkin-tracer/faraday/zipkin-tracer.rb
ykitamura-mdsol: Where is appropriate to define URI? It is just used for BinaryAnnotation's key.

Type::STRING is defined in FinagleThrift. (http://www.rubydoc.info/gems/finagle-thrift/1.4.1/Trace/BinaryAnnotation/Type)

jcarres-mdsol: I'm merging this for now, eventually I feel things will be moved around, good enough for now

PR #55, opened 2016-02-08
adds a logger tracer

jcarres-mdsol:
It is always difficult to know what we would be sending.
This tracer let us see what would be sent. It also may be used if someone wants to send zipkin info to the logs instead of directly to the server.
@jfeltesse-mdsol  @ykitamura-mdsol

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,12 @@ The original tracer, it uses scribe and thrift to send traces.
 
 You need to explicitly install the gem (`gem 'scribe', '~> 0.2.4'`) and set `:scribe_server` in the config.
 
+### Logger
+
+The simplest tracer that does something. It will log all your spans using the logger you pass in the configuration.
+This tracer can be used for debugging purpose (to see what is going to be sent) or to deliver zipkin information into the logs for later retrieval and analysis.
+
+You need to set `:logger` in the configuration.
 
 ### Null
 
--- a/Rakefile
+++ b/Rakefile
@@ -114,8 +114,9 @@ task :benchmark do
   empty_app = EmptyMiddleware.new(nil)
 
   null_configuration = { sample_rate: 1 }
-  json_configuration = null_configuration.merge({ json_api_host: fake_url })
-  scribe_configuration = null_configuration.merge({ scribe_server: fake_url })
+  json_configuration = null_configuration.merge(json_api_host: fake_url)
+  scribe_configuration = null_configuration.merge(scribe_server: fake_url)
+  logger_configuration = null_configuration.merge(logger: logger)
 
   # We create a different faraday middleware per rack middleware below because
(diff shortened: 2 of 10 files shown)


README.md
```diff
@@ -105,6 +105,14 @@ The original tracer, it uses scribe and thrift to send traces.
 
 You need to explicitly install the gem (`gem 'scribe', '~> 0.2.4'`) and set `:scribe_server` in the config.
 
+### Logger
+
+The simplest tracer that does something. It will log all your spans using the logger you pass in the configuration.
+This tracer can be used for debugging purpose (to see what is going to
+be sent) or to deliver zipkin information into the logs for later
+retrieval and analysis.
```
jfeltesse-mdsol: [nits] line breaking is kinda strange

README.md
```diff
@@ -105,6 +105,14 @@ The original tracer, it uses scribe and thrift to send traces.
...
+
+You need to set `:logger` in the configuration.
```
jfeltesse-mdsol: to what value?

lib/zipkin-tracer/config.rb
```diff
@@ -29,6 +30,8 @@ def adapter
         :scribe
       elsif present?(@zookeeper) && RUBY_PLATFORM == 'java'
         :kafka
+      elsif @logger_setup
```
jfeltesse-mdsol: what happens if the user sets `logger: ""` ?

jfeltesse-mdsol:
I'm not sure how to use this, as a user what do I need to set in my config file?

README.md
jcarres-mdsol: it is already explained in the README, the :logger key needs to get an object

lib/zipkin-tracer/config.rb
jcarres-mdsol: that's a bug. We say logger needs an logger object

jcarres-mdsol:
The README already especifies how to use logger, we are just saying that it needs to be set

lib/zipkin-tracer/config.rb
jfeltesse-mdsol: :ok_woman:

jfeltesse-mdsol:
need to amend the commit message: "**t** adds a logger tracer"... :sweat_smile:

Rakefile
```diff
@@ -116,6 +116,7 @@ task :benchmark do
   null_configuration = { sample_rate: 1 }
   json_configuration = null_configuration.merge({ json_api_host: fake_url })
   scribe_configuration = null_configuration.merge({ scribe_server: fake_url })
+  logger_configuration = null_configuration.merge({ logger: logger })
```
jfeltesse-mdsol: [nits] you could do without the braces in there

PR #56, opened 2016-02-10
access to tracer later in faraday so it does not depend on it being set

jcarres-mdsol:
In some weird cases, like some mocking libraries, it can happen that we store a nil into @tracer and then blow up later.
Do not do that, access tracer when needed, it is fast now anyways.

@jfeltesse-mdsol


jcarres-mdsol:
I need to investigate those failures with Ruby 2.1

jfeltesse-mdsol:
LGTM

PR #60, opened 2016-03-01
use hermann 0.27.0.0

jfeltesse-mdsol:
- the new version has been released so no need to use the ref directly.
- not updating changelog or version because it's a dev only change.
- kept it inside the Gemfile and not gemspec so that MRI users don't have to install hermann for nothing when developing.


jfeltesse-mdsol:
@jcarres-mdsol

PR #61, opened 2016-04-05
Allow ZipkinKafkaTracer to use an injected kafka producer instead of hermann

ismith:
Leaves the existing API (provide a `:zookeepers` config option, and a corresponding `Hermann::Producer` is created) alone, but adds an alternative `:producer` option that takes a producer that respond_to?s #push.

### DIFF
--- a/lib/zipkin-tracer/zipkin_kafka_tracer.rb
+++ b/lib/zipkin-tracer/zipkin_kafka_tracer.rb
@@ -1,5 +1,10 @@
-require 'hermann/producer'
-require 'hermann/discovery/zookeeper'
+# IF hermann isn't present, we might be providing another kafka producer
+begin
+  require 'hermann/producer'
+  require 'hermann/discovery/zookeeper'
+rescue LoadError => e
+end
+
 require 'zipkin-tracer/zipkin_tracer_base'
 require 'zipkin-tracer/hostname_resolver'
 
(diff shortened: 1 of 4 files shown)


lib/zipkin-tracer/zipkin_kafka_tracer.rb
```diff
@@ -11,11 +16,23 @@ class ZipkinKafkaTracer < ZipkinTracerBase
...
+        raise ArgumentError, "No (kafka) :producer option (accepting #push) and no :zookeeper option provided."
+      end
       super(options)
     end
 
+    def initialize_hermann_producer(zookeepers)
+      broker_ids = Hermann::Discovery::Zookeeper.new(zookeepers).get_brokers
+      @producer  = Hermann::Producer.new(nil, broker_ids)
+    end
+    private :initialize_hermann_producer
```
jcarres-mdsol: we use `private` section at the bottom of classes instead in this project

spec/lib/zipkin_kafka_tracer_spec.rb
```diff
@@ -26,24 +25,56 @@
...
           expect(tracer.instance_variable_get(:@topic)).to eq topic
         end
       end
+
+      context 'with options including a :producer' do
+        subject { described_class.new(:producer => mock_producer) }
+
+        let(:mock_producer) { double('Kafka producer', :push => true) }
+
+        it 'should use that producer' do
```
jcarres-mdsol: Usually we write in a more active way. Instead of `should use that producer` we do `uses that producer`. Same for other examples.

jfeltesse-mdsol:
pinging @jamescway since he made the original hermann/kafka adapter.

lib/zipkin-tracer/zipkin_kafka_tracer.rb
ismith: Will do.

spec/lib/zipkin_kafka_tracer_spec.rb
ismith: Will do.

ismith:
I'm not sure @jamescway is still interested in this gem.  That aside, though, this change shouldn't affect the API he added; it aims to be a non-breaking change (hence only bumping the minor version).

jcarres-mdsol:
@ismith  fair enough, I'm merging

ismith:
Thanks!

PR #62, opened 2016-05-05
Add a configuration key that actually can make the logger tracer act as such

jcarres-mdsol:
I believe there is no easy (not eval or serialization please) way to pass an object through a configuration file. 
So instead of trying that feat, Adding a "log_tracing" so we use logging if other methods are not available.
Also moved the format to be json which should be easier to parse. 

@jfeltesse-mdsol

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+# 0.18.0
+* Adds the :log_tracing option to explicitly use the logger tracer.
+* Logger classes can not be passed via the configuration file (never worked correctly).
+* Logger tracer logs in JSON format for easy analysis by other tools.
+
+# 0.17.0
+Adds a :producer configuration key as an alternative to Hermann as Kafka client.
+
 # 0.16.0
 * Remove the scribe tracer.
 * Use sucker_punch 2.x. The main feature is the dependency on concurrent-ruby instead of celluloid.
--- a/lib/zipkin-tracer/zipkin_logger_tracer.rb
+++ b/lib/zipkin-tracer/zipkin_logger_tracer.rb
@@ -1,17 +1,24 @@
 require 'zipkin-tracer/zipkin_tracer_base'
 require 'zipkin-tracer/hostname_resolver'
+require 'json'
 
 module Trace
   class ZipkinLoggerTracer < ZipkinTracerBase
+    TRACING_KEY = 'Tracing information'
 
     def initialize(options)
       @logger = options[:logger]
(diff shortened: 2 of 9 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# 0.17.0
```
jfeltesse-mdsol: version 0.17.0 is already tagged and released, see #61 
you should update the changelog accordingly and go for 0.18.0

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
...
+* Added option to explicitly use the logger tracer (log_tracing)
```
jfeltesse-mdsol: [nits] looks like the rest of the changelog is in active tense

lib/zipkin-tracer/zipkin_logger_tracer.rb
```diff
@@ -1,17 +1,24 @@
...
 require 'zipkin-tracer/hostname_resolver'
+require 'json'
 
 module Trace
   class ZipkinLoggerTracer < ZipkinTracerBase
+    TRACING_KEY = 'Tracing information'
 
     def initialize(options)
       @logger = options[:logger]
+      @logger_accepts_data = @logger.class.name.split('::').first == 'Lorekeeper'
```
jfeltesse-mdsol: looks brittle and nowhere in the readme or changelog does it mention lorekeeper. 
also, the ruby way of doing this would be to use `respond_to?` to check the presence of the method you want to use.

jcarres-mdsol: I tried but could not make it work.
Lorekeeper uses method missing magic and `respond_to?` would respond false to any method

jfeltesse-mdsol: well, time to improve lorekeeper then 😉 

it feels wrong to have this ultra specific piece of code referencing a gem many people won't know about, especially without any kind of comment.

CHANGELOG.md
jcarres-mdsol: done

jcarres-mdsol:
@jfeltesse-mdsol pushed requested changes.

jfeltesse-mdsol:
LGTM. Don't have merge powers though.

adriancole:
with great power comes great responsibility.. do you want it? :)

jfeltesse-mdsol:
@adriancole I guess I'm good since Jordi has The Power and we're sitting 2 desks away 😉 
Thanks though!

PR #63, opened 2016-05-12
remove the flags header

jfeltesse-mdsol:
It's not configurable and thus most likely superfluous.

Note: can't nuke all the mentions of the flag because finagle-thrift [expects it](https://github.com/twitter/finagle/blob/develop/finagle-thrift/src/main/ruby/lib/finagle-thrift/trace.rb#L64). Note sure what to do with [this](https://github.com/openzipkin/zipkin-tracer/blob/master/spec/lib/rack/zipkin-tracer_spec.rb#L137-L144) though.

@jcarres-mdsol

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 0.19.0
+* Remove the X-B3-Flags header.
+
 # 0.18.0
 * Adds the :log_tracing option to explicitly use the logger tracer.
 * Logger classes can not be passed via the configuration file (never worked correctly).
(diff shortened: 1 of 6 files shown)


jfeltesse-mdsol:
will update version and changelog, hold on.

adriancole:
well this flag is not needed on the outgoing side unless you are doing something with it (ex you have a way to tell the client to make a debug call)

Inbound, it ought to be read, as you aren't in control of the clients, and they sometimes pass this.

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 0.19.0
+* Remove the X-B3-Flags header.
```
jcarres-mdsol: only on outgoing requests

jfeltesse-mdsol:
@adriancole oh yeah completely skipped that. End of the day PRs are dangerous. Closing for now.

PR #64, opened 2016-05-13
Turn the ZipkinTracer::FaradayHandler::B3_HEADERS constant into a private method

jfeltesse-mdsol:
Title says it all.

No user facing changes.
This is only to make monkey patching of these easier.

@adriancole @jcarres-mdsol


jcarres-mdsol:
nice, thanks

PR #69, opened 2016-06-20
Changed boolean representation of X-B3-Sampled from true/false to 1/0…

### LINKED ISSUE #68: Boolean representation for HTTP_X_B3_SAMPLED incorrect resulting in broken traces
Zipkin-tracer current interprets boolean on HTTP_X_B3_SAMPLED as 'true' or 'false'.

This is out of sync with the header specification and other zipkin tracing interpretations (spring-sleuth/ zipkin-js etc).  Outside of the library implementations themselves the only specification I could find for the headers was here https://twitter.github.io/zipkin/Instrumenting.html#communicating-trace-information

The end effect is that so long as your micro-service ecosystem exclusively uses zipkin-tracer everything is ok but as soon as you start to work in a polyglot ecosystem traces get dropped due to inconsistent interpretation of this header on the inbound and incorrect setting on outbound calls.

Proposal
- Change the interpretation of 'HTTP_X_B3_SAMPLED' to be '0'/'1' inline with specification and other implementations.

singram:
… inline with spec and other implementations.

Resolves #68

### DIFF
--- a/lib/zipkin-tracer/faraday/zipkin-tracer.rb
+++ b/lib/zipkin-tracer/faraday/zipkin-tracer.rb
@@ -14,9 +14,7 @@ module ZipkinTracer
     def call(env)
       trace_id = Trace.id.next_id
       Trace.with_trace_id(trace_id) do
-        b3_headers.each do |method, header|
-          env[:request_headers][header] = trace_id.send(method).to_s
-        end
+        env[:request_headers] = to_http_headers(trace_id)
         if trace_id.sampled?
           trace!(env, trace_id)
         else
@@ -37,6 +35,15 @@ module ZipkinTracer
         sampled: 'X-B3-Sampled',
         flags: 'X-B3-Flags'
       }
+    end
+
+    def to_http_headers(trace_id)
+      headers = {}
+      b3_headers.each do |method, header|
+        headers[header] = trace_id.send(method).to_s
+      end
+      headers['X-B3-Sampled'] = headers['X-B3-Sampled'] == 'true' ? '1' : '0'
+      headers
     end
 
     def trace!(env, trace_id)
--- a/zipkin-tracer.gemspec
+++ b/zipkin-tracer.gemspec
@@ -33,7 +33,7 @@ Gem::Specification.new do |s|
 
   s.add_dependency 'faraday', '~> 0.8'
   s.add_dependency 'finagle-thrift', '~> 1.4.1'
-  s.add_dependency 'rack', '>= 1.0'
+  s.add_dependency 'rack', '< 2.0.0'
   s.add_dependency 'sucker_punch', '~> 2.0'
 
   s.add_development_dependency 'rspec', '~> 3.3'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# 0.19.0
+* Modified X_B3_SAMPLED header to expect & set '1'/'0' as boolean
+representation.  Will also respect a X_B3_SAMPLED header value of 'true'
+on inbound request for backward compaibility.  See http://zipkin.io/pages/instrumenting.html#communicating-trace-information.
+
 # 0.18.2
 * Remove nil parentId from zipkin span payload.
 
--- a/lib/zipkin-tracer/rack/zipkin-tracer.rb
+++ b/lib/zipkin-tracer/rack/zipkin-tracer.rb
@@ -104,7 +104,7 @@ module ZipkinTracer
 
       def should_trace?(parent_trace_sampled)
         if parent_trace_sampled  # A service upstream decided this goes in all the way
-          parent_trace_sampled == 'true'
+          ['1', 'true'].include? parent_trace_sampled
         else
           force_sample? || current_trace_sampled? && !filtered?
         end
--- a/lib/zipkin-tracer/version.rb
+++ b/lib/zipkin-tracer/version.rb
@@ -12,5 +12,5 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 module ZipkinTracer
-  VERSION = '0.18.2'.freeze
+  VERSION = '0.19.0'.freeze
 end
--- a/spec/lib/faraday/zipkin-tracer_spec.rb
+++ b/spec/lib/faraday/zipkin-tracer_spec.rb
@@ -102,7 +102,7 @@ describe ZipkinTracer::FaradayHandler do
         expect(result[:request_headers]['X-B3-ParentSpanId']).to eq('0000000000000003')
         expect(result[:request_headers]['X-B3-SpanId']).not_to eq('0000000000000003')
         expect(result[:request_headers]['X-B3-SpanId']).to match(HEX_REGEX)
-        expect(result[:request_headers]['X-B3-Sampled']).to eq('true')
+        expect(result[:request_headers]['X-B3-Sampled']).to eq('1')
         expect(result[:request_headers]['X-B3-Flags']).to eq('1')
       end
 
@@ -124,7 +124,7 @@ describe ZipkinTracer::FaradayHandler do
         expect(result[:request_headers]['X-B3-TraceId']).to match(HEX_REGEX)
         expect(result[:request_headers]['X-B3-ParentSpanId']).to match(HEX_REGEX)
         expect(result[:request_headers]['X-B3-SpanId']).to match(HEX_REGEX)
-        expect(result[:request_headers]['X-B3-Sampled']).to match(/(true|false)/)
+        expect(result[:request_headers]['X-B3-Sampled']).to match(/(1|0)/)
         expect(result[:request_headers]['X-B3-Flags']).to match(/(1|0)/)
       end
     end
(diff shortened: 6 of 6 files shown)


jcarres-mdsol:
@adriancole  X-B3-Sampled with value 1/0 is correct?

adriancole:
1|0 is more often sent than true|false. Usually, the server will accept either true|false or 1|0

singram:
I'll add a change log entry for this one inline with the other pull request comment.

singram:
@adriancole Is the right approach to converge on a single standard or support both true/false and 1/0?  From my perspective a single agreed representation simplifies expectations, testing & implementation but I'm new to this particular project/ecosystem (thinking the handling of multiple header values (true+1) as you cannot assume what the receiving server will interpret as true).  Either way, given the promise of tracing across polyglot services there appears to be a gap in the documentation covering the http header specification that's inappropriate for swagger.  Perhaps somewhere on the main OpenZipkin readme?  Let me know, happy to take a stab.

adriancole:
I don't like the idea of making true|false a standard for this header.
We've already documented 1|0 here http://zipkin.io/pages/instrumenting.html
and these headers are confusing enough already.

At the moment behavior of "true"|"false" is undefined... there exist
libraries who accept it, but that doesn't mean we should standardize on
that and then incur the busy work of chasing up people who don't accept it
to now accept it. Moreover, it complicates the story about emitting the
header, as now we have 2 options for what is right to emit! Just becomes a
mess.

Bottom-line, if the idea of accepting what was formerly accepted by this
library (true|false) creeps you out, I'd just say don't accept it. The last
thing we need is more complex if statements ported to standard docs or
integration tests.

adriancole:
in case timezone's hinder us.. here's another option. Even if we have
the standard as 1|0, we could add an implementation note to the web
site like this..

Note: Some older tracers set the value "true" or "false" as opposed to
"1" or "0". Some servers accept both variants for transition purposes.

Basically a paragraph break could keep this behavior optional to
support (ie. non-standard), while at the same time document its
existence in case people want to know about it without asking. There
are other places where implementation notes exist, for example
"unknown" as a service name.

singram:
I think we're on the same page here.  I'd like to change the header to support 1|0 in alignment with other implementations and document the change (represented by current branch).  My original problem was that zipkin-tracer was not continuing tracing from other services (e.g. sleuth).  This would correct that and make zipkin-tracer compatible with a wider range of zipkin trace implementations and also track to the specification documentation you highlighted (which I'd missed earlier)

adriancole:
great.. I'm glad my blathering didn't lead to nowhere :)

singram:
LOL not at all.  I think the current pull request lines up with the conversation unless I've got my wires crossed :-)

lib/zipkin-tracer/faraday/zipkin-tracer.rb
```diff
@@ -17,6 +17,7 @@ def call(env)
         b3_headers.each do |method, header|
           env[:request_headers][header] = trace_id.send(method).to_s
         end
+        env[:request_headers]['X-B3-Sampled'] = env[:request_headers]['X-B3-Sampled'] == 'true' ? '1' : '0'
```
jcarres-mdsol: This probably should be in the rack middleware instead if I understand what you are doing.
Aren't you trying to get all incoming `true` to become `1`?

jcarres-mdsol: Oh I see, you need this because the TraceId object itself generate true/false.

Then what I'd do is making a method , maybe in the trace module that takes a TraceId object and returns a TraceId object which true/false has been converted to 1/0

Then use that in both rack and faraday because that ensures that if someone else (older versions of ruby zipkin-tracer or other implementations) send you a true/false you understand them but when you sent, you only sent 1/0

Also I'd up the version to 0.19.0 , even 1.0, some day we have to do it. Because it may break compatibility with other implementations (certainly with older versions of the client)

singram: At that point there would it not be simpler to adjust the inbound test  to be something like
(lib/zipkin-tracer/rack/zipkin-tracer.rb)
`['1','true'].include? parent_trace_sampled`
That would allow the flexibility of the field on the inbound to accept '1'/'true' for compatibility and force '1' on the outbound?  The edge translation between http and the TraceId object is kind of messy.  Perhaps a dedicated object would be a better abstraction?  Something like
class TraceIdHttpHeaderConverter
  def toTraceId
  def toHeaders
end
Naming conventions could probably be a little better but you get the gist.  Let me know what you think.

jcarres-mdsol:
A different object sounds good, keeps the logic isolated and it is very easy to add a comment explaining why it exists.

singram:
@jcarres-mdsol Looking further at this approach.  There's already a Http->obj abstraction in the form of ZipkinEnv.  The work on the Faraday wrapper didn't seem to warrant a separate object so I ended up moving this to a separate method and changing the method->header map to a constant.

lib/zipkin-tracer/faraday/zipkin-tracer.rb
```diff
@@ -29,14 +27,21 @@ def call(env)
...
 
-    def b3_headers
-      {
-        trace_id: 'X-B3-TraceId',
-        parent_id: 'X-B3-ParentSpanId',
-        span_id: 'X-B3-SpanId',
-        sampled: 'X-B3-Sampled',
-        flags: 'X-B3-Flags'
-      }
+    B3_HEADERS = {
```
jcarres-mdsol: The changes are good but can we keep this method as it is instead? 
We have a gem internally that depens on it.

Otherwise I'd merge the pr

singram: @jcarres-mdsol  NP.  Reverted in latest push.

pavolloffay:
@jcarres-mdsol Is this active? I run into the same issue with a polyglot environment.

adriancole:
any way we can pick this back up? the polyglot example is broke on account of it. https://github.com/hawkular/hawkular-apm/tree/master/examples/polyglot-zipkin

singram:
Hi Adrian,
This is good to go from my standpoint, as per the last comments.  Let me
know if there's anything I need to do to get this unstuck.
- Stuart

adriancole:
looks like a rebase is all that's needed, right @jcarres-mdsol ?

zipkin-tracer.gemspec
```diff
@@ -33,7 +33,7 @@ Gem::Specification.new do |s|
 
   s.add_dependency 'faraday', '~> 0.8'
   s.add_dependency 'finagle-thrift', '~> 1.4.1'
-  s.add_dependency 'rack', '>= 1.0'
+  s.add_dependency 'rack', '< 2.0.0'
```
jcarres-mdsol: Rails 5 uses Rack > 2.0 .
Why change this?

jcarres-mdsol:
Only one comment, then just need to rebase and it is good to merge!

singram:
Np.  I'll take a look at this later today.

jcarres-mdsol:
So, I've been thinking about this and going back to the comments of Adrian there, I think this PR does two things:
- When discovered a true/false force it to 1/0
- Accept both true/false and 1/0

The problem with the first point is that a service with this tracer will break a system which is full of tracers which only accept true/false. In particular if you have rails services, you will be forced to upgrade all of them to 0.19 or 0.18 will break the chain.

So this PR becomes more complicated but to avoid breaking any system the best thing is to propagate whatever you get. If you get true/false, propagate those, else propagate 1/0
If a new one needs to be created you can default to 1/0 so all your system will be 1/0 from origin to end.

Else, this would be a 1.0 kind of release and a difficult one to upgrade to as everyone needs to upgrade at the same time, as the contract between services have changed.

adriancole:
could you solve this with a flag/option? currently it is unusable in a
system that is written correctly.

adriancole:
For example, that way you can upgrade the libraries and eventually
change the option in such a site. You could also "double-check" ex. it
is very odd to propagate an id and also have sampled=0/false. You
could use a heuristic that double-checks until library updates
converge.

adriancole:
How about this. When we receive 'X-B3-Sampled', we interpret it as a boolean (eg either "true" or 1 means True) but leave the actual value alone, and propagate it in whatever ruby client the same way as it was received. For example, we pass it opaquely as it was sent to us. This should solve the brown field.

Then, the problem is simplified to the first emission of 'X-B3-Sampled'. In this case, we could have a configuration option, which says to write either true|false or 1|0 when sampling a trace. This setting could be set to `write_sampled_as=true_false` to retain historical behavior. At some point, all the libraries will be updated, then they can change to `write_sampled_as=1_0`.

New installations would default to `write_sampled_as=1_0` which makes them compatible with all the other propagators.

singram:
Sounds reasonable.
However I could see a case where you do want to transform a value to
correctly work with another service.
Having control over both the pass through representation and initial sample
value would handle all possible use cases.
Thoughts around the default stance(s)?  It seems to me that the
write_sampled_as=1_0 is more standards compliant at this time.

JordiPolo:
I've reimplemented this PR in a backwards compatible way, can you take a look at #79 ?

jcarres-mdsol:
closing this as #79 provides similar functionality. Thanks for the inspiration and pushing for this!

PR #71, opened 2016-08-23
Removed assumption about hostnames starting with letters 

matthill:
Docker containers have hostnames that start with numbers.  It's based on the container hash which you can't control.  The "ipv4" address was being sent as (for example) 465cc3595a43 which caused Zipkin to fail when indexing.  This fix removes the regex check that assumes the hostnames must start with letters.

### DIFF
--- a/lib/zipkin-tracer/hostname_resolver.rb
+++ b/lib/zipkin-tracer/hostname_resolver.rb
@@ -8,9 +8,7 @@ module ZipkinTracer
 
       each_annotation(spans) do |annotation|
         hostname = annotation.host.ipv4
-        if hostname.to_s =~ /\A[a-zA-Z]/ # hostnames start with letters, else we already have an IP
-          annotation.host.ipv4 = host_to_ip[hostname]
-        end
+        annotation.host.ipv4 = host_to_ip[hostname]
       end
     end
 


jcarres-mdsol:
Interesting, I think that check is there to protect against resolving IPs. maybe should be checking if it is not an IP address instead.
Also failing test

adriancole:
https://github.com/openzipkin/zipkin-tracer/issues/75 seems related. cc @pavolloffay

ipv4 is an ip literal in json, not a hostname. similarly for ipv6

```
      ipv4:
        type: string
        description: 'The text representation of a IPv4 address associated with this endpoint. Ex. 192.168.99.100'
      ipv6:
        type: string
        description: 'The text representation of a IPv6 address associated with this endpoint. Ex. 2001:db8::c001'
```
https://github.com/openzipkin/zipkin-api/blob/master/zipkin-api.yaml#L205

adriancole:
ps this is a dependency of this pretty amazing looking zipkin demo https://github.com/hawkular/hawkular-apm/tree/master/examples/polyglot-zipkin

adriancole:
yeah I think the goal here is to make sure this isn't an ipv4 or ipv6 literal. Is there a library for that? else I could try and make one.

jcarres-mdsol:
Implemented in #76 . Closing

PR #73, opened 2016-09-13
Standardize key of binary annotation representing http URL

pavolloffay:
Make key of http URL binary annotation standardized.
See https://github.com/openzipkin/zipkin/blob/master/zipkin/src/main/java/zipkin/TraceKeys.java#L79

@jcarres-mdsol could you please review?


jcarres-mdsol:
Good idea, but I've been looking at this and it seems what it really means its http.path which is also standard. I'll open another PR changing it to that.

jcarres-mdsol:
Implemented as `http.path` in #77  . Thanks for bringing this up!

pavolloffay:
Thanks for solving it :+1: .

PR #76, opened 2016-11-02
Ensures ip addresses for all hostnames are resolved, solves issue with docker hostnames

jcarres-mdsol:
Sorry for sitting on this so long.
I think it is a better implementation than #71  and should solve #75 
@matthil @adriancole  @ajanques @ssteeg-mdsol  @ykitamura-mdsol

### DIFF
--- a/lib/zipkin-tracer/hostname_resolver.rb
+++ b/lib/zipkin-tracer/hostname_resolver.rb
@@ -1,3 +1,5 @@
+require 'resolv'
+
 module ZipkinTracer
   # Resolves hostnames in the endpoints of the annotations.
   # Resolving hostnames is a very expensive operation. We want to store them raw in the main thread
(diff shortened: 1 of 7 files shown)


lib/zipkin-tracer/hostname_resolver.rb
```diff
@@ -19,6 +21,13 @@ def spans_with_ips(spans)
     LOCALHOST_I32 = 0x7f000001.freeze
     IP_FIELD = 3
 
+    def resolved_ip_address?(ip_string)
+      # When the ip_format is string, we will match with one of these two regexp
+      # When the ip_format is :i32 (used by kafka), we just check the string is a number
+      !!((ip_string =~ Resolv::IPv4::Regex) || (ip_string =~ Resolv::IPv6::Regex)) ||
```
adriancole: if we found an ipv6 stuffed in the ipv4 field, we should probably move it to the ipv6 one :)

jcarres-mdsol: This is in preparation when we support ipv6. We will need a different pr for that

ykitamura-mdsol: you can use`Regexp.union`, if you like
```ruby
ip_string =~ Regexp.union(Resolv::IPv4::Regex, Resolv::IPv6::Regex)
```

ykitamura-mdsol:
Looks good to me

PR #77, opened 2016-11-02
uses http.path instead of http.uri for paths

jcarres-mdsol:
All in the title
@adriancole  @ykitamura-mdsol  @ssteeg-mdsol


ykitamura-mdsol:
Looks good to me

PR #79, opened 2016-11-04
adds support for 1/0 as values of the sampled header

JordiPolo:
Reimplementing #69 

After this PR zipkin-tracer will:
- Accept either true/false or 1/0
- Propagate whatever sampled value it gets
- If it is the first service in the trace (no value for sampled) it will user true/false or 1/0 depending on a configuration option

@adriancole  @singram

This PR still needs changing of Changelog, document the new configuration, etc but opening early for comments.

### DIFF
--- a/lib/zipkin-tracer/config.rb
+++ b/lib/zipkin-tracer/config.rb
@@ -35,6 +35,10 @@ module ZipkinTracer
       @log_tracing       = config[:log_tracing]
       # When set to false, it uses 1/0 in the 'X-B3-Sampled' header, else uses true/false
       @sampled_as_boolean = config[:sampled_as_boolean] || DEFAULTS[:sampled_as_boolean]
+      # The current default is true for compatibility but services are encouraged to move on.
+      if @sampled_as_boolean
+        @logger && @logger.warn("Using a boolean in the Sampled header is deprecated. Consider setting sampled_as_boolean to false")
+      end
     end
 
     def adapter
--- a/lib/zipkin-tracer/config.rb
+++ b/lib/zipkin-tracer/config.rb
@@ -6,20 +6,35 @@ module ZipkinTracer
   class Config
     attr_reader :service_name, :service_port, :json_api_host,
       :zookeeper, :sample_rate, :logger, :log_tracing,
-      :annotate_plugin, :filter_plugin, :whitelist_plugin
+      :annotate_plugin, :filter_plugin, :whitelist_plugin,
+      :sampled_as_boolean
 
     def initialize(app, config_hash)
       config = config_hash || Application.config(app)
+      # The name of the current service
       @service_name      = config[:service_name]
+      # The port where the current service is running
       @service_port      = config[:service_port]      || DEFAULTS[:service_port]
+      # The address of the Zipkin server which we will send traces to
       @json_api_host     = config[:json_api_host]
+      # Zookeeper information
       @zookeeper         = config[:zookeeper]
+      # Percentage of traces which by default this service traces (as float, 1.0 means 100%)
       @sample_rate       = config[:sample_rate]       || DEFAULTS[:sample_rate]
+      # A block of code which can be called to do extra annotations of traces
       @annotate_plugin   = config[:annotate_plugin]   # call for trace annotation
       @filter_plugin     = config[:filter_plugin]     # skip tracing if returns false
       @whitelist_plugin  = config[:whitelist_plugin]  # force sampling if returns true
+      # A block of code which can be called to skip traces. Skip tracing if returns false
+      @filter_plugin     = config[:filter_plugin]
+      # A block of code which can be called to force sampling. Forces sampling if returns true
+      @whitelist_plugin  = config[:whitelist_plugin]
       @logger            = Application.logger
       @log_tracing       = config[:log_tracing]       # Was the logger in fact setup by the client?
+      # Was the logger in fact setup by the client?
+      @log_tracing       = config[:log_tracing]
+      # When set to false, it uses 1/0 in the 'X-B3-Sampled' header, else uses true/false
+      @sampled_as_boolean = config[:sampled_as_boolean] || DEFAULTS[:sampled_as_boolean]
     end
 
     def adapter
@@ -38,7 +53,8 @@ module ZipkinTracer
 
     DEFAULTS = {
       sample_rate: 0.1,
-      service_port: 80
+      service_port: 80,
+      sampled_as_boolean: true
     }
 
     def present?(str)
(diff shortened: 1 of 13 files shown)


lib/zipkin-tracer/config.rb
```diff
@@ -38,7 +53,8 @@ def adapter
 
     DEFAULTS = {
       sample_rate: 0.1,
-      service_port: 80
+      service_port: 80,
+      sampled_as_boolean: true
```
adriancole: since it is spec to use 1|0 shouldn't we default to that?

adriancole:
this is a lot less code than I was expecting 👍

lib/zipkin-tracer/config.rb
JordiPolo: on the other hand it would change the current behavior without warning.
Maybe leave this default but log a deprecation warning, then after a while we switch the default.

adriancole:
>
> the other hand it would change the current behavior without warning.
> Maybe leave this default but log a deprecation warning, then after a while
> we switch the default.
>
SGTM

jcarres-mdsol:
@singram This reimplements the changes you need in a backward compatible way. 
Can you test if it works for you?

adriancole:
@singram on holiday? :) if so, please attach photo of a beach!

singram:
Sorry all, have been swamped recently, unfortunately no beach in sight @adriancole .  
Will verify the newer branch this weekend.

singram:
@jcarres-mdsol  Ran into issues with the new Thrift version as a result of the Faraday version bump.  This was with both docker images for  ruby:2.3.2-slim &  ruby:2.3.2-alpine.  Missing development dependency looked to be the issue.  Using the stock 'ruby:2.3.2' as a fallback worked.  

With the `sampled_as_boolean` config property in place this all worked.  Thanks

jcarres-mdsol:
if faraday 0.9 creates issues I guess if fair enough as they are still in 0.x , it kind of sucks as much of the projects I work with and I suspect others work with are moving to 0.9 already. I really do not want to lock into the older 0.8 branch.
I do not know what's the differences among ruby docker images, I have no idea why they would give problems. Can you give us more details?
Is this version then workable for you?

singram:
Hi @JordiPolo, the branch works, ideally I'd like to use the smaller images but for my use case it's not critical.  I'm absolutely agreeing the thrift/faraday version should move forwards.

Where I'm coming from this that zipkin is likely to be deployed as part of a distributed system solution.  As such it's likely in a production environment that slimer docker images are advantageous.  It would be helpful to document any work around for the *-slim and *-alpine images to ease user adoption.

All that being said, I've not had the time to look properly (:)) but it appears to be a simple development dependency missing.  Here's the details from a bundle install I got with the newer version

```
Step 8 : RUN gem install bundler     && bundle install
 ---> Running in 6af5ad7255af
Successfully installed bundler-1.13.6
1 gem installed
Fetching gem metadata from https://rubygems.org/.........
Fetching version metadata from https://rubygems.org/..
Fetching dependency metadata from https://rubygems.org/.
Fetching https://github.com/JordiPolo/zipkin-tracer.git
Installing concurrent-ruby 1.0.2
Installing multipart-post 2.0.0
Installing thrift 0.9.3.0 with native extensions
Installing rack 1.6.4
Installing tilt 2.0.5
Using bundler 1.13.6
Installing sucker_punch 2.0.2
Installing faraday 0.10.0
Gem::Ext::BuildError: ERROR: Failed to build gem native extension.

    current directory: /usr/local/bundle/gems/thrift-0.9.3.0/ext
/usr/local/bin/ruby -r ./siteconf20161123-9-1q48fi2.rb extconf.rb
checking for strlcpy() in string.h... *** extconf.rb failed ***
Could not create Makefile due to some reason, probably lack of necessary
libraries and/or headers.  Check the mkmf.log file for more details.  You may
need configuration options.

Provided configuration options:
	--with-opt-dir
	--without-opt-dir
	--with-opt-include
	--without-opt-include=${opt-dir}/include
	--with-opt-lib
	--without-opt-lib=${opt-dir}/lib
	--with-make-prog
	--without-make-prog
	--srcdir=.
	--curdir
	--ruby=/usr/local/bin/$(RUBY_BASE_NAME)
/usr/local/lib/ruby/2.3.0/mkmf.rb:456:in `try_do': The compiler failed to
generate an executable file. (RuntimeError)
You have to install development tools first.
	from /usr/local/lib/ruby/2.3.0/mkmf.rb:541:in `try_link0'
	from /usr/local/lib/ruby/2.3.0/mkmf.rb:556:in `try_link'
	from /usr/local/lib/ruby/2.3.0/mkmf.rb:765:in `try_func'
	from /usr/local/lib/ruby/2.3.0/mkmf.rb:1051:in `block in have_func'
	from /usr/local/lib/ruby/2.3.0/mkmf.rb:942:in `block in checking_for'
	from /usr/local/lib/ruby/2.3.0/mkmf.rb:350:in `block (2 levels) in postpone'
	from /usr/local/lib/ruby/2.3.0/mkmf.rb:320:in `open'
	from /usr/local/lib/ruby/2.3.0/mkmf.rb:350:in `block in postpone'
	from /usr/local/lib/ruby/2.3.0/mkmf.rb:320:in `open'
	from /usr/local/lib/ruby/2.3.0/mkmf.rb:346:in `postpone'
	from /usr/local/lib/ruby/2.3.0/mkmf.rb:941:in `checking_for'
	from /usr/local/lib/ruby/2.3.0/mkmf.rb:1050:in `have_func'
	from extconf.rb:31:in `<main>'

To see why this extension failed to compile, please check the mkmf.log which can
be found here:

  /usr/local/bundle/extensions/x86_64-linux/2.3.0-static/thrift-0.9.3.0/mkmf.log

extconf failed, exit code 1

Gem files will remain installed in /usr/local/bundle/gems/thrift-0.9.3.0 for
inspection.
Results logged to
/usr/local/bundle/extensions/x86_64-linux/2.3.0-static/thrift-0.9.3.0/gem_make.out

An error occurred while installing thrift (0.9.3.0), and Bundler cannot
continue.
Make sure that `gem install thrift -v '0.9.3.0'` succeeds before bundling.
```

jcarres-mdsol:
I think the problem is unrelated to Faraday.
The problem is that the thrift gem has C code in it and when installing it's trying to compile it. The slim images probably do not have the to do so.

jcarres-mdsol:
I'm merging this for now as it seems to solve one of your problems.
I do not think we will be able to solve the problem with the containers unless we stop depending on `thrift` which is in itself a worthy goal, probably. But please open an issue about that problem so we can keep track of it.

I am going to merge this and close related PRs/issues.

georgeu2000:
I have the following config (in Ruby): 

zipkin_config = { service_name: '...', service_port: settings.port, sampled_as_boolean: false }

But get warning:

`W, [2016-12-16T17:15:07.989034 #24443]  WARN -- : Using a boolean in the Sampled header is deprecated. Consider setting sampled_as_boolean to false`

zipkin-tracer 0.19.1

jcarres-mdsol:
I'll look into it

PR #80, opened 2016-11-07
Name the local span by its name

ys:
This PR allows us to use the name passed to `trace_local_component` instead of the `lc` constant that serves for the binary annotation.

Right now all the local traces created are named `lc`.
This makes it harder to debug local traces.
We have the `lc` binary annotation to categorize this.
According to https://github.com/openzipkin/zipkin/issues/808 this is enough for a local span.

## before

image

## after

image

### DIFF
--- a/lib/zipkin-tracer/trace_client.rb
+++ b/lib/zipkin-tracer/trace_client.rb
@@ -17,7 +17,7 @@ module ZipkinTracer
       result = nil
       if @trace_id.sampled?
         Trace.with_trace_id(@trace_id) do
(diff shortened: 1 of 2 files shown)


ys:
@JordiPolo @adriancole anything I can do to see this land?

JordiPolo:
@adriancole is there any reason we need to annotate local components `lc`? I can not remember if there was a rationale here.

If case Adrian says this change is cool, then you would need to update the version and the Changelog also so we can release a new gem.

ys:
It still has the binary annotation `lc` just the name can be more explicit.

adriancole:
In general, the local component value isn't used as a span name, but that's still better than the span name "lc"

👍

PR #81, opened 2016-11-10
Add a noop flush! in null tracer

ys:
I am instrumenting background jobs and with sidekiq, I have to flush the queue myself at the end of a job.

This will ensure that all the tracers have the `flush!` method implemented.
So that whenever it is traced or not, the tracer reacts similarly.


jcarres-mdsol:
Good change. Thanks!

PR #83, opened 2016-11-17
Add basic sidekiq middleware

ys:
This adds a way to trace background jobs in Sidekiq.

This lacks documentation and tests but is opened early to have a feel of the need of the community for this.

This is used this way.

```
options = {
  service_port:  443,
  service_name:  service_name,
  sample_rate:   ENV.fetch("ZIPKIN_SAMPLE_RATE", 0.1).to_f,
  json_api_host: ENV.fetch("ZIPKIN_API_HOST", "https://zipkin.io")
}

Sidekiq.configure_server do |config|
  config.server_middleware do |chain|
    chain.add ZipkinTracer::Sidekiq::Middleware, options
  end
end
```

### DIFF
--- a/lib/zipkin-tracer/sidekiq/middleware.rb
+++ b/lib/zipkin-tracer/sidekiq/middleware.rb
@@ -1,0 +1,36 @@
+module ZipkinTracer
+  module Sidekiq
+    class Middleware
+      def initialize(app, config = nil)
+        @app = app
+        @config = Config.new(app, config).freeze
+        @tracer = TracerFactory.new.tracer(@config)
+      end
+
+      def sample?
+        rand < @config.sample_rate
+      end
+
+      def call(worker, item, _queue)
+        id = Trace.generate_id
+        trace_id = Trace::TraceId.new(id, nil, id, sample?, ::Trace::Flags::EMPTY)
+
+        result = nil
+        klass = item["wrapped".freeze] || worker.class.to_s
+        ::Trace.with_trace_id(trace_id) do
+          if sample?
+            @tracer.with_new_span(trace_id, klass) do |span|
+              span.record("mr")
+              result = yield
+              span.record("ms")
+            end
+          else
+            result = yield
+          end
+        end
+        ::Trace.tracer.flush! if ::Trace.tracer.respond_to?(:flush!)
+        result
+      end
+    end
+  end
+end


lib/zipkin-tracer/sidekiq/middleware.rb
```diff
@@ -0,0 +1,36 @@
...
+      def call(worker, item, _queue)
+        id = Trace.generate_id
+        trace_id = Trace::TraceId.new(id, nil, id, sample?, ::Trace::Flags::EMPTY)
+
+        result = nil
+        klass = item["wrapped".freeze] || worker.class.to_s
+        ::Trace.with_trace_id(trace_id) do
+          if sample?
+            @tracer.with_new_span(trace_id, klass) do |span|
+              span.record("mr")
```
adriancole: ooo messaging span!

ys: I haven't seen much examples so I kind of guessed too:)

jcarres-mdsol:
I've never used sidekiq myself but if it is used to do some async work on behalf of a request (send an email for instance), won't it be useful to pass the traceid of the current request to the job?

Or because of the timings implications that may have, it's something we want to avoid @adriancole ?

atmos:
> won't it be useful to pass the traceid of the current request to the job?

This was my thinking as well. Have a convention that sidekiq can look for an inject if applicable.

adriancole:
I think the yield part might have thrown me off (as it looked like a function was being invoked synchronously in the same thread). I think sidekiq has various functionality in async which means we might have different patterns depending on if something leaves the process or not.

There is indeed a use case for the current trace (and span etc) in [messaging spans](https://github.com/openzipkin/zipkin/issues/1243)

For example, there are zipkin instrumentation out there that propagate almost exactly the same way across messages as they do http requests (by adding trace, span etc as message headers).

In that case, you'd add "ms" when you propagate the span onto the message, and "mr" when you pop it off the other side. It is probably best to create a child span for whatever processing occurs after that.

You can use "ms", "mr" similarly in local processing, just in any case take note that it can significantly affect the span's timescale if the activity is delayed orders of magnitude. In such cases, it is more like another trace. We've not formalized [linked traces](https://github.com/openzipkin/zipkin/issues/1244), but one could start by adding a binary annotation of the producing trace, in these circumstances.

ys:
I think @adriancole idea of getting the source span id as a binary annotation makes sense.

But maybe it is ok to have long traces for this... I guess it depends on how fast sidekiq or similar just process the events.

adriancole:
simplest is long traces.. can always revisit. For example, in spring we are
starting with long traces.

jcarres-mdsol:
I'm kind of a little lost, is the PR as it is good or there are specific changes to be made?

adriancole:
@jcarres-mdsol I think the report timestamp and duration (without necessarily adding annotations), the comment https://github.com/openzipkin/zipkin-ruby/pull/83#discussion_r88766885 can resolve without adding "ms" and "mr" (which are for different purposes)

basically a test like.

* start new local span, with a binary annotation of "lc"-> "sidekiq"
  * this span wraps the job processing and when reported includes span.timestamp and duration

ex. 

```ruby
# the following works like a benchmark, recording the timestamp and duration of the block represented by yield.
@tracer.with_new_span(trace_id, klass) do |span|
   result = yield
end
```
The difference from the current PR, is that we aren't adding "ms" and "mr" annotations, as they imply [remote one-way](https://github.com/openzipkin/zipkin/issues/1243) messaging (and also propagation of trace identifiers)

In any case, an example json or screenshot will help show this works!

ys:
Thanks for the comments, I will come back with something that is more clear.

ys:
I will come with a new PR later based on the discussions

Issue #84, opened 2016-11-18
Forcing all paths to get GET causes problems


keithmgould:
Currently in the `Application.rb` file the gem uses the [recognize_path](https://github.com/rails/rails/blob/master/actionpack/lib/action_dispatch/routing/route_set.rb#L728) method.  No second parameter is sent in. This means all paths are forced to be GET.

Of course that is not the case.

Wondering if I'm missing something, or if we need a PR to pass in a second parameter (a hash) with a `:method` key?

PR #85, opened 2016-11-18
passes the http verb to the recognize_path method

keithmgould:
Currently the http verb is not passed to the recognize_path method. This means that all paths are interpreted as GET (see [here](https://github.com/rails/rails/blob/master/actionpack/lib/action_dispatch/routing/route_set.rb#L728)).

We are currently getting false negatives due to this problem.

This PR fixes by passing in the http verb.

Due to the method (recognize_path) being part of Rails (not this gem) we had to continue to only stub the method, so I could not make a true "failing test."

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 0.18.6
+* Passes HTTP Method to recognize_path
+
 # 0.18.5
 * `NullTracer` has a noop `flush!` method.
 * Spans from `local_component_span` will be named according to `local_component_value` over `lc`.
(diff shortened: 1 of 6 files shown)


keithmgould:
Adresses: https://github.com/openzipkin/zipkin-ruby/issues/84

jcarres-mdsol:
Cool, good changes. 
The CI failed on JRuby but seems a transient error to me, try pushing again.
Also, you'll need to update the version and the Changelog so we can release a new version

adriancole:
ps opened #86 in case it helps w/jruby flake

keithmgould:
thanks @jcarres-mdsol.

bumped version and edited change log.

jcarres-mdsol:
Merging, thanks!

PR #86, opened 2016-11-19
Uses Java 8 when running JRuby

adriancole:
hopefully this can help with build flakes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,12 @@
 language: ruby
+jdk:
+  - oraclejdk8
 rvm:
   - 2.3.0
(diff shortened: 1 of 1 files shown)


jcarres-mdsol:
Any idea why these problems happen? Should we have officially supported java versions?

adriancole:
I haven't tracked down anything in particular, just that occasionally OS
packages change and trip off things. Using JRE 8 is a sensible first step
before debugging other things (as JRE 7 was end of life last year.

jcarres-mdsol:
well, EOL means no love anymore. Merging.

Issue #87, opened 2016-11-19
Parent span ID header should be optional


rgueldem:
We are using the zipkin tracer rack middleware and it doesn't properly handle requests that have a trace ID, span ID and sampled flag but no parent span ID. According to the documentation, the parent span ID should be fully optional.

https://github.com/openzipkin/zipkin-ruby/blob/master/lib/zipkin-tracer/rack/zipkin-tracer.rb#L24 lists the `X-B3-ParentSpanID` header as required, which will fail the `called_with_zipkin_headers?` check. Moving it to the optional headers should fix it, but will break the order in `trace_parameters`.

adriancole:
PS the context of this is that there are debugging tools such as the chrome
plugin who control the trace and span id of the root by propagating them
(usually they also set flags=1 and sampled=1 to ensure the trace is saved).
In this case, there's no parent span id. So when a parent span id wasn't
propagated, we indicate this is the root span with an externally
provisioned ID.

rgueldem:
If it helps, this is the fix I use locally:
```diff
diff --git a/lib/zipkin-tracer/rack/zipkin-tracer.rb b/lib/zipkin-tracer/rack/zipkin-tracer.rb
index 7552522..d7af5c5 100644
--- a/lib/zipkin-tracer/rack/zipkin-tracer.rb
+++ b/lib/zipkin-tracer/rack/zipkin-tracer.rb
@@ -21,8 +21,8 @@ module ZipkinTracer
   # This middleware reads Zipkin headers from the request and sets/creates a Trace.id usable by the rest of the app
   # It will also send the trace to the Zipkin service using one of the methods configured.
   class RackHandler
-    B3_REQUIRED_HEADERS = %w[HTTP_X_B3_TRACEID HTTP_X_B3_PARENTSPANID HTTP_X_B3_SPANID HTTP_X_B3_SAMPLED].freeze
-    B3_OPT_HEADERS = %w[HTTP_X_B3_FLAGS].freeze
+    B3_REQUIRED_HEADERS = %w[HTTP_X_B3_TRACEID HTTP_X_B3_SPANID HTTP_X_B3_SAMPLED].freeze
+    B3_OPT_HEADERS = %w[HTTP_X_B3_PARENTSPANID HTTP_X_B3_FLAGS].freeze
 
     def initialize(app, config = nil)
       @app = app
@@ -79,13 +79,13 @@ def trace_id(default_flags = Trace::Flags::EMPTY)
[...]

PR #90, opened 2016-11-27
limits the required zipkin headers to x_b3_trace_id and x_b3_span_id

jcarres-mdsol:
Changes tracer to behave as per zipkin spec, the required headers are only the ones for trace_id and span_id.
Should solve #87 

I've just discovered that having a class under private in Ruby does not make it private, you need to use `private_constant` instead which I haven't seen in my life and I guess other developers will be surprised to see it. So I've moved the previously_thought_to_be_private class into its own file and made the modifications there.

@ykitamura-mdsol  @jfeltesse-mdsol  @ssteeg-mdsol  @adriancole  @rgueldem

### DIFF
--- a/lib/zipkin-tracer/rack/zipkin_env.rb
+++ b/lib/zipkin-tracer/rack/zipkin_env.rb
@@ -1,0 +1,72 @@
+module ZipkinTracer
+  # This class manages the Zipkin related information in the Rack environment.
+  # It is only used by the rack middleware.
+  class ZipkinEnv
+    attr_reader :env
+
+    def initialize(env, config)
+      @env    = env
+      @config = config
+    end
+
+    def trace_id(default_flags = Trace::Flags::EMPTY)
+      trace_id, span_id, parent_span_id = retrieve_or_generate_ids
+      sampled = sampled_header_value(@env['HTTP_X_B3_SAMPLED'])
+      flags = (@env['HTTP_X_B3_FLAGS'] || default_flags).to_i
+      Trace::TraceId.new(trace_id, parent_span_id, span_id, sampled, flags)
+    end
+
+    def called_with_zipkin_headers?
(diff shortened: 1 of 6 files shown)


lib/zipkin-tracer/rack/zipkin_env.rb
```diff
@@ -0,0 +1,72 @@
+module ZipkinTracer
+  # This class manages the Zipkin related information in the Rack environment.
+  # It is only used by the rack middleware.
+  class ZipkinEnv
+    attr_reader :env
+
+    def initialize(env, config)
+      @env    = env
```
jfeltesse-mdsol: [nits] what does rubocop think of this?

spec/lib/rack/zipkin-tracer_spec.rb
```diff
@@ -55,7 +55,7 @@ def expect_host(host)
 
   context 'Zipkin headers are passed to the middleware' do
     subject { middleware(app) }
-    let(:env) {mock_env(',', ZipkinTracer::RackHandler::B3_REQUIRED_HEADERS.map {|a| Hash[a, 1] }.inject(:merge))}
+    let(:env) {mock_env(',', ZipkinTracer::ZipkinEnv::B3_REQUIRED_HEADERS.map {|a| Hash[a, 1] }.inject(:merge))}
```
jfeltesse-mdsol: [nits] lack of spaces in there

lib/zipkin-tracer/rack/zipkin_env.rb
jcarres-mdsol: no complains, seems to be ok to follow that style

ykitamura-mdsol:
looks good to me

adriancole:
the code looks better each time. Nice one! and thanks for the continued progress on this!

PR #92, opened 2016-11-28
Adds a LICENSE file and deletes individual references to license

jcarres-mdsol:
Changes as in the title.
The original files did have Copyright Twitter. Given the files have changed so much probably nothing is left from those times, I do not know what to do with that @adriancole  please advise.

@jfeltesse-mdsol  @ssteeg-mdsol  @ykitamura-mdsol  FYI.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,202 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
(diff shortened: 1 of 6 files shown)


LICENSE
```diff
@@ -0,0 +1,202 @@
...
+   APPENDIX: How to apply the Apache License to your work.
+
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
```
jfeltesse-mdsol: I know it says otherwise in that appendix but can't we just keep the section below and call it a day??
I see literally nowhere people copy/pasting the entire license but I see the section below all the time. I'm not a lawyer but It clearly states and links to the license itself, isn't that good enough?

jfeltesse-mdsol:
If you have access to the settings of the repo you can additionally set the license type so that it appears on the top page of the repo.

LICENSE
adriancole: we don't have to pretend we know which file includes which copyrights. A common punt is to simply say: `Copyright 2015-2016 The OpenZipkin Authors`

jcarres-mdsol:
I have no access to settings, probably @adriancole has

adriancole:
@jcarres-mdsol try now :P

jcarres-mdsol:
it seems it is more about github detecting the file: https://help.github.com/articles/adding-a-license-to-a-repository/

LICENSE
jcarres-mdsol: I just copied this file from the openzipkin/zipkin repo.
I am good about adding a `NOTICE` file. I really dislike the header in the files because what I look for there is some comment to help me understand the code, it's frustrating to find a bunch of text which is no helpful at all.

adriancole:
currently, we don't have any requirement to put license headers in any
files (we aren't a part of some foundation with such a mandate). I've
certainly done projects that have no license headers, except a license file
in the root dir. It is also possible to add them as a part of the
distribution (ex adding them to source only in the gem).

TL;DR; no problem from my side if you'd like to scrub the headers, just add
a NOTICE with the names of the files which originally had one. Maybe also
add a CONTRIBUTING file or similar that implies that pull requests imply
ASL licensing.

LICENSE
jcarres-mdsol: I've added a super small notice file

NOTICE.md
```diff
@@ -0,0 +1,3 @@
+This repository was originally developed by Twitter and released under Apache license.
```
adriancole: LG

PR #93, opened 2016-11-30
Add bin. annotation http.path for every request

pavolloffay:
Even though incoming request comes from a traced service it is not guaranteed that `http.path|url|uri` was added at a client side.

@jcarres-mdsol could you please review?

### DIFF
--- a/lib/zipkin-tracer/rack/zipkin-tracer.rb
+++ b/lib/zipkin-tracer/rack/zipkin-tracer.rb
@@ -53,7 +53,7 @@ module ZipkinTracer
 
     def trace!(span, zipkin_env, &block)
       # if called by a service, the caller already added the information
-      trace_request_information(span, zipkin_env.env) unless zipkin_env.called_with_zipkin_headers?
+      trace_request_information(span, zipkin_env.env)
       span.record(Trace::Annotation::SERVER_RECV)
       span.record('whitelisted') if zipkin_env.force_sample?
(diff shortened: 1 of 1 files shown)


jcarres-mdsol:
@adriancole  knows more about all the other implementations out there.
The two I know more about C# and Ruby will always set the path in the client so we avoid setting it again in the server.
In what environment is this failing for you?

lib/zipkin-tracer/rack/zipkin-tracer.rb
```diff
@@ -53,7 +53,7 @@ def annotate_plugin(env, status, response_headers, response_body)
 
     def trace!(span, zipkin_env, &block)
       # if called by a service, the caller already added the information
-      trace_request_information(span, zipkin_env.env) unless zipkin_env.called_with_zipkin_headers?
```
adriancole: in a shared span (which is the current model), this will duplicate the entry (once for the client and again for the server). For many, making this change will clutter the span view and also add data with no value unless there's some proxy going on. I'm working in the java (Brave) codebase to make something like this opt-in vs required for all users.

when/if we move to a single-host model, I'd agree with this change regardless, but for now I do think it will clutter most people's span detail screen.
https://github.com/openzipkin/zipkin/issues/939#issuecomment-264076909

pavolloffay:
@jcarres-mdsol @adriancole thanks for feedback. I'm closing it.

PR #94, opened 2016-12-16
fixes faraday middleware and add new convenient classes

jcarres-mdsol:
The faraday middleware created ids and cleaned up after him which is good,  except when there were no ids to begin with.
This fixes the issue and allows to always clean up after him.
Also added more classes because I like classes

@ykitamura-mdsol  @adriancole  @jfeltesse-mdsol  @ssteeg-mdsol

### DIFF
--- a/lib/zipkin-tracer/config.rb
+++ b/lib/zipkin-tracer/config.rb
@@ -39,6 +39,7 @@ module ZipkinTracer
       if @sampled_as_boolean
         @logger && @logger.warn("Using a boolean in the Sampled header is deprecated. Consider setting sampled_as_boolean to false")
       end
+      Trace.sample_rate = @sample_rate
     end
 
     def adapter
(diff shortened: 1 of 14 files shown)


lib/zipkin-tracer/config.rb
```diff
@@ -39,6 +39,7 @@ def initialize(app, config_hash)
       if @sampled_as_boolean
         @logger && @logger.warn("Using a boolean in the Sampled header is deprecated. Consider setting sampled_as_boolean to false")
       end
+      Trace.sample_rate = @sample_rate
```
ykitamura-mdsol: 👍

spec/lib/config_spec.rb
```diff
@@ -6,13 +6,17 @@ module ZipkinTracer
       allow(Application).to receive(:logger).and_return(Logger.new(nil))
     end
     [:service_name, :service_port, :json_api_host,
-      :zookeeper, :sample_rate, :log_tracing,
+      :zookeeper, :log_tracing,
       :annotate_plugin, :filter_plugin, :whitelist_plugin].each do |method|
-      it "can set and read configuration values for #{method}" do
+      it 'can set and read configuration values for #{method}' do
```
ykitamura-mdsol: should be `" ~ "`

jcarres-mdsol:
Still internal details, give it a new couple of versions to mature, just in case.

PR #97, opened 2017-02-23
Add an Excon middleware

tbalthazar:
Add an Excon middleware that allows to trace outgoing web requests created by [Excon](https://github.com/excon/excon), the same way `zipkin-ruby` allows to trace outboing web requests created by Faraday.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# 0.20.2
+# 0.21.0
 * Added an Excon middleware
 
 # 0.20.1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 0.20.2
+* Added an Excon middleware
+
 # 0.20.1
 * Bugfix: Properly handle the `sampled_as_boolean` configuration option
 
--- a/lib/zipkin-tracer/excon/zipkin-tracer.rb
+++ b/lib/zipkin-tracer/excon/zipkin-tracer.rb
@@ -4,65 +4,70 @@ require 'uri'
 require 'excon'
 
 module ZipkinTracer
-    class ExconHandler < Excon::Middleware::Base
-      def initialize(stack)
-        @stack = stack
-        # Excon does not currently provide a way to parameterize middlewares.
-        @service_name = ENV.fetch('ZIPKIN_SERVICE_NAME', 'unknown-service')
-      end
+  class ExconHandler < Excon::Middleware::Base
+    def initialize(stack)
+      @stack = stack
+      # Excon does not currently provide a way to parameterize middlewares.
+      @service_name = ENV.fetch('ZIPKIN_SERVICE_NAME', 'unknown-service')
+    end
 
-      def error_call(datum)
-        # do stuff
-        puts "error call"
-        @stack.error_call(datum)
-      end
+    def error_call(datum)
+      # do stuff
+      puts "error call"
+      @stack.error_call(datum)
+    end
 
-      def request_call(datum)
-        puts "request call"
-        byebug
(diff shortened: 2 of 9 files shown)


lib/zipkin-tracer/excon/zipkin-tracer.rb
```diff
@@ -0,0 +1,83 @@
+require 'finagle-thrift/trace'
+require 'finagle-thrift/tracer'
+require 'uri'
+require 'excon'
+
+module ZipkinTracer
+  class ExconHandler < Excon::Middleware::Base
+    def initialize(stack)
```
jcarres-mdsol: Maybe `_` instead of stack so there is no unused param

lib/zipkin-tracer/excon/zipkin-tracer.rb
```diff
@@ -0,0 +1,83 @@
...
+      super
+    end
+
+    def error_call(datum)
+      super(datum)
+    end
+
+    def request_call(datum)
```
jcarres-mdsol: `datum` is not very specific, doesn't Excon use a more specific term for these?

jcarres-mdsol: Same comment for other `datum` in this file.

jcarres-mdsol:
Probably you would want to rescue errors , for instance parsing of the URL can fail

lib/zipkin-tracer/excon/zipkin-tracer.rb
ys: This is sadly the term they give in excon... That couldn't be more generic:D

lib/zipkin-tracer/excon/zipkin-tracer.rb
tbalthazar: Done!

tbalthazar:
@jcarres-mdsol Thanks for your remarks, I added a commit. Please let me know if you have any other suggestion.

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 0.20.2
+* Added an Excon middleware
```
adriancole: heh should this be 0.21? seems new middleware is more than a patch

lib/zipkin-tracer/excon/zipkin-tracer.rb
```diff
@@ -0,0 +1,85 @@
...
+          datum[:headers][header] = trace_id.send(method).to_s
+        end
+
+        trace!(datum, trace_id) if trace_id.sampled?
+      end
+
+      super(datum)
+    end
+
+    def response_call(datum)
```
adriancole: if there's some sort of unhandled exception, you can add a binary annotation of type string with key "error" with the cause. This will make it show up in the UI in red.

CHANGELOG.md
tbalthazar: Makes sense, thanks for your feedback!

zipkin-tracer.gemspec
```diff
@@ -22,6 +22,7 @@ Gem::Specification.new do |s|
 
   s.add_dependency 'faraday', '~> 0.8'
   s.add_dependency 'finagle-thrift', '~> 1.4.2'
+  s.add_dependency 'excon', '~> 0.53'
```
JordiPolo: could this be a development dependency? It will be a tough sell to make everyone to depend on excon

tbalthazar: Good point, makes sense to me!

jcarres-mdsol:
Thanks for the addition, merging it in !

adriancole:
✨

PR #100, opened 2017-03-17
Add tests for response status tracing in Excon middleware

tbalthazar:
Test the response status is correctly traced.  Also guard against nil
response/status, which might happen if the middleware is inserted into
Excon before the Excon::Middleware::ResponseParser.

### DIFF
--- a/lib/zipkin-tracer/excon/zipkin-tracer.rb
+++ b/lib/zipkin-tracer/excon/zipkin-tracer.rb
@@ -70,8 +70,7 @@ module ZipkinTracer
     end
 
     def response_status(datum)
-      return nil unless datum[:response] && datum[:response][:status]
-      return datum[:response][:status].to_s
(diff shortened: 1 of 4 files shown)


lib/zipkin-tracer/excon/zipkin-tracer.rb
```diff
@@ -64,6 +69,11 @@ def service_name(datum, default)
       datum.fetch(:zipkin_service_name, default)
     end
 
+    def response_status(datum)
+      return nil unless datum[:response] && datum[:response][:status]
```
jcarres-mdsol: In this case I'd write this slightly more simply and in the positive:
```
if datum[:response] && datum[:response][:status]
  datum[:response][:status].to_s
else
  nil
end

tbalthazar: Good point!

PR #101, opened 2017-03-24
Prevents raising an error if the tracer is not set in Faraday and Excon

jcarres-mdsol:
Also reworked the specs to test the case when we add headers but do not trace 
@adriancole


jcarres-mdsol:
@ykitamura-mdsol  @ssteeg-mdsol

ykitamura-mdsol:
Tested locally and it worked.
Looks good to me.

PR #103, opened 2017-04-21
Add the `:record_on_server_receive` configuration option

jfeltesse-mdsol:
This adds a new configuration option to force the sending of path info on server receive.
Usually it is not needed because it's provided by the client but we have a use case here that requires that data to be sent (in a nutshell some service processing zipkin data to compute metrics).

@jcarres-mdsol @adriancole

### DIFF
--- a/lib/zipkin-tracer/config.rb
+++ b/lib/zipkin-tracer/config.rb
@@ -40,8 +40,7 @@ module ZipkinTracer
         @logger && @logger.warn("Using a boolean in the Sampled header is deprecated. Consider setting sampled_as_boolean to false")
       end
       # Record the given tags on server receive, even if the zipkin headers were present in the incoming request?
-      tags = config[:record_on_server_receive]
-      @record_on_server_receive = present?(tags) ? parse_record_on_server_receive(tags) : nil
+      @record_on_server_receive = parse_tags(config[:record_on_server_receive])
 
       Trace.sample_rate = @sample_rate
     end
@@ -66,19 +65,17 @@ module ZipkinTracer
       sampled_as_boolean: true
     }
 
-    def parse_record_on_server_receive(tag_names)
+    def parse_tags(tag_names)
+      return {} unless present?(tag_names)
       names = tag_names.split(",").map(&:strip)
-      tags = {}
-      (ZipkinTracer::RackHandler::DEFAULT_SERVER_RECV_TAGS.keys & names).each do |name|
+      (ZipkinTracer::RackHandler::DEFAULT_SERVER_RECV_TAGS.keys & names).each_with_object({}) do |name, tags|
         tags[name] = ZipkinTracer::RackHandler::DEFAULT_SERVER_RECV_TAGS[name]
       end
-      tags.empty? ? nil : tags
     end
 
     def present?(str)
       return false if str.nil?
       !!(/\A[[:space:]]*\z/ !~ str)
     end
-
   end
 end
--- a/lib/zipkin-tracer/config.rb
+++ b/lib/zipkin-tracer/config.rb
@@ -1,5 +1,6 @@
 require 'logger'
 require 'zipkin-tracer/application'
+require 'zipkin-tracer/rack/zipkin-tracer'
 
 module ZipkinTracer
   # Configuration of this gem. It reads the configuration and provides default values
@@ -7,7 +8,7 @@ module ZipkinTracer
     attr_reader :service_name, :service_port, :json_api_host,
       :zookeeper, :sample_rate, :logger, :log_tracing,
       :annotate_plugin, :filter_plugin, :whitelist_plugin,
-      :sampled_as_boolean
+      :sampled_as_boolean, :record_on_server_receive
 
     def initialize(app, config_hash)
       config = config_hash || Application.config(app)
@@ -30,7 +31,6 @@ module ZipkinTracer
       # A block of code which can be called to force sampling. Forces sampling if returns true
       @whitelist_plugin  = config[:whitelist_plugin]
       @logger            = Application.logger
-      @log_tracing       = config[:log_tracing]       # Was the logger in fact setup by the client?
       # Was the logger in fact setup by the client?
       @log_tracing       = config[:log_tracing]
       # When set to false, it uses 1/0 in the 'X-B3-Sampled' header, else uses true/false
@@ -39,6 +39,10 @@ module ZipkinTracer
       if @sampled_as_boolean
         @logger && @logger.warn("Using a boolean in the Sampled header is deprecated. Consider setting sampled_as_boolean to false")
       end
+      # Record the given tags on server receive, even if the zipkin headers were present in the incoming request?
+      tags = config[:record_on_server_receive]
+      @record_on_server_receive = present?(tags) ? parse_record_on_server_receive(tags) : nil
+
       Trace.sample_rate = @sample_rate
     end
 
@@ -61,6 +65,15 @@ module ZipkinTracer
       service_port: 80,
       sampled_as_boolean: true
     }
+
+    def parse_record_on_server_receive(tag_names)
+      names = tag_names.split(",").map(&:strip)
+      tags = {}
+      (ZipkinTracer::RackHandler::DEFAULT_SERVER_RECV_TAGS.keys & names).each do |name|
+        tags[name] = ZipkinTracer::RackHandler::DEFAULT_SERVER_RECV_TAGS[name]
+      end
+      tags.empty? ? nil : tags
+    end
 
     def present?(str)
       return false if str.nil?
--- a/lib/zipkin-tracer/rack/zipkin-tracer.rb
+++ b/lib/zipkin-tracer/rack/zipkin-tracer.rb
@@ -50,8 +50,8 @@ module ZipkinTracer
     def trace!(span, zipkin_env, &block)
       # if the request comes from a non zipkin-enabled source record the default tags
       tags = DEFAULT_SERVER_RECV_TAGS unless zipkin_env.called_with_zipkin_headers?
-      # if the user specified tags to record on server recv, use these no matter what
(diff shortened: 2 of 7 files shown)


lib/zipkin-tracer/config.rb
```diff
@@ -39,6 +38,9 @@ def initialize(app, config_hash)
       if @sampled_as_boolean
         @logger && @logger.warn("Using a boolean in the Sampled header is deprecated. Consider setting sampled_as_boolean to false")
       end
+      # Record the path info on server receive, even if the zipkin headers were present in the incoming request?
```
adriancole: I've thought about this a little elsewhere and kindof like the keys approach, where you specify the keys you would like to have added to the span. I'm not sure if this config is shared between client and server or not. If it isn't you could imagine a set that includes "http.uri" etc.

jcarres-mdsol: You mean the configuration file would have a list of keys to be added on each span for both the client and the server?

adriancole:
>
> You mean the configuration file would have a list of keys to be added on
> each span for both the client and the server?
>
yep. this is particularly neat with http.url vs http.uri, but also when
people add new keys like "requestId" as then you can declaratively decide
whether to add it even if the code is always there

jfeltesse-mdsol:
@adriancole talking with @jcarres-mdsol offline we came to the conclusion that I'm going to change that configuration option to something more generic that would allow you to specify keys that are recognized by the client/server, such as http.path, in CSV style.
Note that right now it's only that one which is "duplicated", the client doesn't send http.uri for instance.
I'll amend this PR and comment back.

adriancole:
sounds great. I'm excited to see where this goes. we might use similar
approach in other tracers.

jfeltesse-mdsol:
changed the option to `:record_on_server_receive` which accepts CSV style tag names, whitelisting them though. It only supports `http.path` for now, not sure what would make sense to add...

lib/zipkin-tracer/config.rb
```diff
@@ -39,6 +39,10 @@ def initialize(app, config_hash)
       if @sampled_as_boolean
         @logger && @logger.warn("Using a boolean in the Sampled header is deprecated. Consider setting sampled_as_boolean to false")
       end
+      # Record the given tags on server receive, even if the zipkin headers were present in the incoming request?
+      tags = config[:record_on_server_receive]
+      @record_on_server_receive = present?(tags) ? parse_record_on_server_receive(tags) : nil
```
jcarres-mdsol: I would do `{}` instead of `nil` and later you do not need to check for `nil` just do `.each` on the hash

lib/zipkin-tracer/rack/zipkin-tracer.rb
```diff
@@ -8,6 +9,14 @@ module ZipkinTracer
   # This middleware reads Zipkin headers from the request and sets/creates a Trace.id usable by the rest of the app
   # It will also send the trace to the Zipkin service using one of the methods configured.
   class RackHandler
+    # the following constants are defined only from rack 1.6
```
jcarres-mdsol: could we depend on `rack > 1.6` ?

jfeltesse-mdsol: Not really if we want to keep a reasonable backwards-compatibility.
For instance actionpack 4.1.x depends on rack ~> 1.5.2 (https://rubygems.org/gems/actionpack/versions/4.1.16)

lib/zipkin-tracer/config.rb
```diff
@@ -62,6 +66,15 @@ def adapter
       sampled_as_boolean: true
     }
 
+    def parse_record_on_server_receive(tag_names)
```
adriancole: I don't know what's idiomatic, but seems like it would be easier to see this if this function were in the rack class (ex tag filter was passed to it vs moving the logic here). For example, it would make it easier to see if you change "parse_record_on_server_receive" to "parse_tags" and no meaning is loss. I was looking to see if any tags were added on server send and needed to scroll to the rack handler file to find that out..

jfeltesse-mdsol: Not sure I get it, here I'm parsing the value passed in the config once and for all, just like the other config options. I don't think that parsing whatever was passed at config time belongs in the rack class since it has nothing to do with middleware stuff.

lib/zipkin-tracer/config.rb
jfeltesse-mdsol: the reason I went this way is that the standard case is the user doesn't have this option set _and_ the service is called by another zipkin-enabled one. In this 99% of the time case, we don't need to do anything at all, e.g.

- `zipkin_env.called_with_zipkin_headers?` is true 
- `@config.record_on_server_receive` is falsy 
=> `tags` var is nil 
=> `trace_request_information` doesn't do anything

If I go the default empty hash route here I'll either have to check for nil and empty or I'll have to create a new hash object every time just so that `.each` doesn't blow.

jcarres-mdsol: The empty hash would be made only here in config and then stay empty, then you do not need to check for it, maybe you'll need to call merge on it and each.

On the other hand `nil` brings no meaning, given that something blows up and it is a nil, it is difficult to know the intention of that data

jfeltesse-mdsol: something like this? https://github.com/openzipkin/zipkin-ruby/pull/103/commits/c41332a45f30f6178f76908c77006b5ab300c26d

lib/zipkin-tracer/rack/zipkin-tracer.rb
```diff
@@ -50,8 +50,8 @@ def annotate_plugin(env, status, response_headers, response_body)
     def trace!(span, zipkin_env, &block)
       # if the request comes from a non zipkin-enabled source record the default tags
       tags = DEFAULT_SERVER_RECV_TAGS unless zipkin_env.called_with_zipkin_headers?
-      # if the user specified tags to record on server recv, use these no matter what
-      tags = @config.record_on_server_receive if @config.record_on_server_receive
+      # if the user specified tags to record on server receive, use these no matter what
```
adriancole: nit spelling receive

adriancole:
sorry I misunderstood. thanks for clearing!

lib/zipkin-tracer/rack/zipkin-tracer.rb
jfeltesse-mdsol: ? Am I missing something? Any spelling issue with the following?
"if the user specified tags to record on server receive, use these no matter what"

adriancole: weird.. I thought I saw "recieve" here! Ignore me .. I'm not helping I think :P

lib/zipkin-tracer/rack/zipkin-tracer.rb
```diff
@@ -50,8 +50,8 @@ def annotate_plugin(env, status, response_headers, response_body)
...
+      tags = @config.record_on_server_receive unless @config.record_on_server_receive.empty?
```
jcarres-mdsol: why do this? just let the empty hash reach `trace_request_information`

jfeltesse-mdsol: true, that was in line with the old tags can be nil code but it doesn't make sense anymore

jfeltesse-mdsol: ah, nope, can't do this because of the case where we'd been called by a non zipkin enabled service, `tags` would be using the `DEFAULT_SERVER_RECV_TAGS`. Getting rid of the check here would overwrite this and make it empty...
And I can't use that `DEFAULT_SERVER_RECV_TAGS` as the config's default altogether because we still need it to be empty in the 99% case where we're called with zipkin headers and we don't want the http.path to be recorded too.
In a nutshell, we need the empty default config while having an alternative default for the case of no zipkin headers in the request.
Makes sense?

lib/zipkin-tracer/rack/zipkin-tracer.rb
```diff
@@ -31,16 +40,20 @@ def call(env)
...
 
     def annotate_plugin(env, status, response_headers, response_body)
       @config.annotate_plugin.call(env, status, response_headers, response_body) if @config.annotate_plugin
     end
 
     def trace!(span, zipkin_env, &block)
-      # if called by a service, the caller already added the information
-      trace_request_information(span, zipkin_env.env) unless zipkin_env.called_with_zipkin_headers?
+      # if the request comes from a non zipkin-enabled source record the default tags
+      tags = DEFAULT_SERVER_RECV_TAGS unless zipkin_env.called_with_zipkin_headers?
```
jcarres-mdsol: maybe do these things inside `trace_request_information` ?

lib/zipkin-tracer/rack/zipkin-tracer.rb
```diff
@@ -49,8 +62,11 @@ def trace!(span, zipkin_env, &block)
       span.record(Trace::Annotation::SERVER_SEND)
     end
 
-    def trace_request_information(span, env)
-      span.record_tag(Trace::BinaryAnnotation::PATH, env['PATH_INFO'])
+    def trace_request_information(span, env, tags)
+      # tags is nil if we've been called by a zipkin-enabled service and the user hasn't
+      # specified tags to record on server receive.
+      return if tags.nil?
```
jcarres-mdsol: Maybe...
```ruby
tags =  if !@config.record_on_server_receive.empty?
  @config.record_on_server_receive
else if !zipkin_env.called_with_zipkin_headers?
  DEFAULT_SERVER_RECV_TAGS
else
  {}
end
```

jfeltesse-mdsol: updated following your recommendation but left the `return if tags.nil?` check so as not to needlessly create empty hash objects and call .each on them in what is going to be the 99% use case.

PR #104, opened 2017-05-04
add sidekiq middleware for tracing sidekiq jobs

ayrat555:
I added [sidekiq middleware](https://github.com/mperham/sidekiq/wiki/Middleware) for tracing background jobs

The tracing can be turned on by adding ZipkinTracer::Sidekiq::Middleware to your sidekiq middleware chain

```
zipkin_tracer_config = {
  service_name: 'service',
  service_port: 3000,
  json_api_host: 'http://zipkin.io',
  traceable_workers: [:MyWorker, :MyWorker2],
  sample_rate: 0.5
}

Sidekiq.configure_server do |config|
  config.server_middleware do |chain|
    chain.add ::ZipkinTracer::Sidekiq::Middleware, zipkin_tracer_config
  end
end
```
You can specify the workers that you want to trace with **traceable_workers** config option. If you want all your workers to be traced pass [:all] to traceable_workers option (traceable_workers: [:all])

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 0.22.1
+* Add sidekiq worker tracing.
+
 # 0.22.0
 * Add the `:record_on_server_receive` configuration option.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# 0.22.1
+# 0.23.0
 * Add sidekiq worker tracing.
 
 # 0.22.0
--- a/lib/zipkin-tracer/sidekiq/middleware.rb
+++ b/lib/zipkin-tracer/sidekiq/middleware.rb
@@ -1,0 +1,47 @@
+module ZipkinTracer
+  module Sidekiq
+    class Middleware
+      attr_reader :config, :tracer, :traceable_workers
+
+      def initialize(config)
+        @config = Config.new(nil, config).freeze
+        @tracer = TracerFactory.new.tracer(@config)
+        @traceable_workers = config.fetch(:traceable_workers, [])
+      end
+
+      def call(worker, job, queue, &block)
+        return block.call unless traceable_worker?(worker)
+
+        trace(worker, job, queue, &block)
+      end
+
+      private
+
+      def traceable_worker?(worker)
+        traceable_workers.include?(:all) || traceable_workers.include?(worker_name(worker))
+      end
+
+      def trace(worker, job, queue, &block)
+        trace_id = TraceGenerator.new.next_trace_id
+        span_name = worker_name(worker)
+
+        result = TraceContainer.with_trace_id(trace_id) do
+          if trace_id.sampled?
+            tracer.with_new_span(trace_id, span_name) do
+              result = block.call
+            end
+          else
+            result = block.call
+          end
+        end
+
+        tracer.flush!
+        result
+      end
+
+      def worker_name(worker)
+        worker.class.to_s.to_sym
+      end
+    end
+  end
+end
--- a/README.md
+++ b/README.md
@@ -56,6 +56,28 @@ Note that supplying the service name for the destination service is optional;
 the tracing will default to a service name derived from the first section of the destination URL (e.g. 'service.example.com' => 'service').
 
 
+### Tracing Sidekiq workers
+
+Sidekiq tracing can be turned on by adding ZipkinTracer::Sidekiq::Middleware to your sidekiq middleware chain:
+
+```ruby
+zipkin_tracer_config = {
+  service_name: 'service',
+  service_port: 3000,
+  json_api_host: 'http://zipkin.io',
+  traceable_workers: [:MyWorker, :MyWorker2],
+  sample_rate: 0.5
+}
+
+Sidekiq.configure_server do |config|
+  config.server_middleware do |chain|
+    chain.add ZipkinTracer::Sidekiq::Middleware, zipkin_tracer_config
+  end
+end
+```
+
+By default workers aren't traced. You can specify the workers that you want to trace with traceable_workers config option. If you want all your workers to be traced pass [:all] to traceable_workers option (traceable_workers: [:all]).
+
 ### Local tracing
 
 `ZipkinTracer::TraceClient` provides an API to record local traces in your application.
--- a/lib/zipkin-tracer.rb
+++ b/lib/zipkin-tracer.rb
@@ -1,6 +1,7 @@
 require 'base64' #Bug in finagle. They should be requiring this: finagle-thrift-1.4.1/lib/finagle-thrift/tracer.rb:115
 require 'zipkin-tracer/trace'
 require 'zipkin-tracer/rack/zipkin-tracer'
+require 'zipkin-tracer/sidekiq/middleware'
 require 'zipkin-tracer/trace_client'
 require 'zipkin-tracer/trace_container'
 require 'zipkin-tracer/trace_generator'
--- a/lib/zipkin-tracer/version.rb
+++ b/lib/zipkin-tracer/version.rb
@@ -1,3 +1,3 @@
 module ZipkinTracer
-  VERSION = '0.22.0'.freeze
+  VERSION = '0.22.1'.freeze
 end
--- a/lib/zipkin-tracer/version.rb
+++ b/lib/zipkin-tracer/version.rb
@@ -1,3 +1,3 @@
 module ZipkinTracer
-  VERSION = '0.22.1'.freeze
+  VERSION = '0.23.0'.freeze
 end
--- a/spec/lib/sidekiq/middleware_spec.rb
+++ b/spec/lib/sidekiq/middleware_spec.rb
@@ -1,0 +1,33 @@
+require 'spec_helper'
+
+describe ZipkinTracer::Sidekiq::Middleware do
+
+  describe 'call' do
+    subject { described_class.new(config) }
+    let(:job) { "test_job" }
+    let(:queue) { "test_queue" }
+
+    context 'with traceable_workers config option' do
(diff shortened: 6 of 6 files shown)


jcarres-mdsol:
Thanks!
We'll review next week @ykitamura-mdsol  @jfeltesse-mdsol

lib/zipkin-tracer/sidekiq/middleware.rb
```diff
@@ -0,0 +1,52 @@
...
+            result = block.call
+          end
+        end
+
+        tracer.flush!
+        result
+      end
+
+      def sample?
+        rand < config.sample_rate
```
jfeltesse-mdsol: isn't there a way to use the usual `trace_id.sampled?` method instead of re-implementing this logic again?
see https://github.com/openzipkin/zipkin-ruby/blob/master/lib/zipkin-tracer/rack/zipkin-tracer.rb#L28-L30

If not, I guess a comment explaining why would be helpful.

jfeltesse-mdsol:
btw the version, changelog and readme need to be updated as well.

adriancole:
I didn't see reference to #83 by @ys did you look at that?

lib/zipkin-tracer/sidekiq/middleware.rb
ys: We probably want the `sampled?` from trace id otherwise it might not trace a trace that was traced or the opposite.

ys:
@adriancole Yeah haven't looked at the diffs, I am right now trying to integrate it correctly in our stack with the Sidekiq client and server so that we have beautiful traces. And also across some kafka:D

I will be back with that somewhere in the next month or so I guess

lib/zipkin-tracer/sidekiq/middleware.rb
ayrat555: @jfeltesse-mdsol @ys yes, my bad. fixed

ayrat555:
@jfeltesse-mdsol I updated the version, changelog and readme
@adriancole yes, I saw this PR. but unfortunately, it got nowhere

ys:
@ayrat555 I am working on it. 
It went nowhere because there are much more than just the server side. 
I want to propagate the trace from the point where the sidekiq job is enqueued and not consider this as solely a new trace.

ayrat555:
@ys great. but sometimes you need a new trace for a sidekiq worker (I need this right now for the project I'm working on)

ayrat555:
@adriancole @jfeltesse-mdsol can we merge this feature now and improve it in the future? I think it's useful addition to zipkin-ruby functionality because sidekiq is extensively used in rails community. Also currently our company is migrating to zipkin so there will be more contributions from me or my colleagues.

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 0.22.1
```
adriancole: new things are likely a new minor 0.23.0

adriancole:
@ayrat555 can you find someone else who will use this and have them vouch for this? While generically this is a good idea I don't know if the pattern used here is what others want or not. best way is to have folks answer that directly.

if you don't know anyone, you can search github for others doing this and mention them here.. usually there's some ad-hoc tracing code for most things

tank-bohr:
@adriancole Almost everything can be done with ad-hoc workarounds. One can easily implement this middleware in his project. This proposal is for consistency sake. Rack and excon extensions are here. That's a good thing to have a sidekiq middleware here as well. I don't understand why don't you discuss the code but discuss the intention. The Intention is obviously worthwhile. If the code is fine I think you should just accept it

adriancole:
@tank-bohr unless code is reusable, it is tech debt. This is a very important thing to keep in mind in any open source project. Putting code in place that others cannot use is worse than no code in place. There's a nice blog about this in general here https://blog.codinghorror.com/rule-of-three/

No one should take offence at such a question, it isn't personal in any way, as you are right it has nothing to do with this implementation! As far as I can tell at least 8 people are not only interested in sidekiq, but this implementation of it. So, rule-of-three is clearly met imho!

I didn't mean to ruffle feathers, sorry if I did. I just want folks to be able to support eachother as there's ton's of abandoned code out there.

tank-bohr:
@adriancole Thank you! I'm agree with you, useless code is harmful. I'm sorry I was offensive. I didn't mean to

adriancole:
all good I took no offense, just wanted to explain my mysterious ways.
thanks for letting me :P

tank-bohr:
> I want to propagate the trace from the point where the sidekiq job is enqueued and not consider this as solely a new trace.

@ys I think it's a very bad idea. Sidekiq job is totally async stuff. It should be a new trace
You'll get something like that

```
controller action
controller action
enqueue sidekiq job
controller action
sidekiq job
controller action
sidekiq job
sidekiq job
controller action
```

You can't specify an order. Sidekiq works in parallel processes. It can be enqueued with delay. Your traces will not be beautiful they will be entangled

ys:
I think I disagree with you here as it depends on the usage.
For example, we are using Sidekiq to process webhooks that were posted to our app.
I want to be able in some ways to trace it from the controller to the last service it touches.
if you passes span ids and so on, you get strange traces but they make sense. 
It is like using sidekiq as a local message broker.

ys:
And by the way starting by saying something is bad without knowing the context is not a way to engage in a conversation with someone :)

tank-bohr:
I'm sorry for being so rude. In my humble opinion, context doesn't matter in this case. I consider it's a bad idea in general. I believe it might be useful in some narrow cases. But in general, it's very confusing. Sidekiq is not sequential by design. You can't wait a sidekiq job to be finished. It's in contradiction to zipkin usage pattern, I think. Thanks

ys:
@ayrat555 if you wanna take a look at how I am trying to do it, it looks pretty similar to you right now so I think it is great:D

https://github.com/openzipkin/zipkin-ruby/compare/master...ys:sidekiq

stellacotton:
@adriancole- stepping out of the instrumentation weeds a bit, have there been any decisions on where Zipkin as a community is going with handling async/fanout from the UI level? 

It sounds like there are two valid problem statements that @ys and @tank-bohr are addressing here:
1) How do we represent request times accurately? If async jobs are included in the trace, then it becomes much harder to debug latency issues because the trace timing includes non-blocking, async jobs.
2) How do we represent the path a request is taking through the system? If we treat async jobs as a new trace, then it becomes much harder to debug failures in the system because the trace isn't aggregated as a single unit.

Is linked traces is where we ultimately want to end up? 
 https://github.com/openzipkin/zipkin/issues/1244

And in that world, it sounds like @ayrat555's implementation makes sense because it would represent latency more accurately but also help us visualize failures in the request path?

adriancole:
The only thing currently implemented is one-way single producer single
consumer aka half duplex via "cs" "sr". People are using this approach
today.

Linked traces are interesting but not quite understood yet. Meeting next
month on this particularly w/the census team who have experience.

On 19 May 2017 05:22, "Stella Cotton" <notifications@github.com> wrote:

@adriancole- stepping out of the instrumentation weeds a bit, have there
been any decisions on where Zipkin as a community is going with handling
async/fanout from the UI level?

It sounds like there are two valid problem statements that @ys
<https://github.com/ys> and @tank-bohr <https://github.com/tank-bohr> are
addressing here:

   1. How do we represent request times accurately? If async jobs are
   included in the trace, then it becomes much harder to debug latency issues
   because the trace timing includes non-blocking, async jobs.
   2. How do we represent the path a request is taking through the system?
   If we treat async jobs as a new trace, then it becomes much harder to debug
   failures in the system because the trace isn't aggregated as a single unit.

Is linked traces is where we ultimately want to end up?
openzipkin/zipkin#1244 <https://github.com/openzipkin/zipkin/issues/1244>

And in that world, it sounds like @ayrat555 <https://github.com/ayrat555>'s
implementation makes sense because it would represent latency more
accurately but also help us visualize failures in the request path?

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/openzipkin/zipkin-ruby/pull/104#issuecomment-302545503>,
or mute the thread

stellacotton:
@adriancole got it, thanks for the update!

jcarres-mdsol:
@ys  I like your approach, but why use job["tracing_information"]?

Here, another similar approach: https://gist.github.com/jcarres-mdsol/d36dd0986094914c0a6a1cce5b347e52

You'd use this as:
```
ZipkinTracer::TracedJob.trace(name: 'my_background_thing') do
 connection = Faraday.new do |builder|
    builder.use ZipkinTracer::FaradayHandler
    builder.adapter Faraday.default_adapter
  end

end
```

`ZipkinTracer::TracedJob.trace` could surround the whole job or only an interaction from some loop, etc as needed

ys:
@jcarres-mdsol so the information part is to pass the trace information from sidekiq client side to server side.
So it survives the Redis serialization.
That's it.
I want to be able to trace the whole client server process.

ayrat555:
@jcarres-mdsol Actually in my project I implemented easy DSL that traces code as needed
https://gist.github.com/ayrat555/17094411038f1801db7bd8194079d7c7

I can make PR with it including tests

jcarres-mdsol:
@ayrat555 @ys   This seems ready to merge, maybe just fix the changelog conflict and we can merge and release a new version

ayrat555:
@jcarres-mdsol done

adriancole:
PS looping back from discussion last month. https://github.com/openzipkin/zipkin/issues/1654 for multiple consumers and how it impacts instrumentation

chaintng:
I think i agree with @ys that each workers are supposed to have traceId.

I saw sidekiq worker as another entity application (microservices) and i would like to trace time between each microservices, this is what zipkin was designed for.

Is this pull request have an option for that?

PR #105, opened 2017-05-10
Fix excon span duration bug

ys:
Because of the way the middlewares in Excon works, we have to split the `with_new_span` call to `start_span` in the request part of the middleware and `end_span` in the response part.

This also changes the specs to ensure we call start and end directly. And add those methods to the null tracer.

### DIFF
--- a/spec/lib/excon/zipkin-tracer_spec.rb
+++ b/spec/lib/excon/zipkin-tracer_spec.rb
@@ -52,6 +52,7 @@ describe ZipkinTracer::ExconHandler do
   end
 
   it 'has a trace with correct duration' do
+    request_duration_in_seconds = 1
     Timecop.freeze
     Trace.tracer = Trace::NullTracer.new
     ::Trace.sample_rate = 1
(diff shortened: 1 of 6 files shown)


adriancole:
it there a way to add a unit test that shows the duration bug no longer exists? code looks sensible to me

ys:
We wrote the middleware.. I just tested with one app that has it and it is fixed. 
Time tests are always hard I will see.

ys:
@adriancole Found a way, tell what do you think about it.

spec/lib/excon/zipkin-tracer_spec.rb
```diff
@@ -51,6 +51,34 @@ def process(body, url, headers = {})
...
 
+  it 'has a trace with correct duration' do
+    Trace.tracer = Trace::NullTracer.new
+    ::Trace.sample_rate = 1
+    trace_id = ::Trace::TraceId.new(1, 2, 3, true, ::Trace::Flags::DEBUG)
+    url = 'https://www.example.com'
+    allow(::Trace).to receive(:default_endpoint)
+      .and_return(::Trace::Endpoint.new('127.0.0.1', '80', 'example.com'))
+    stub_request(:post, url)
+      .to_return(body: lambda { |request| sleep 1; '' })
```
adriancole: clever. I think eventually the orchestration around this could be refactored. The key fact lost here is how we end up with 1_000_000. I think it is because of the sleep. Might be worth making the argument to sleep a variable like duration_seconds, and multiply that in the  later assertion will make more sense.

ys:
Thanks to you, should I bump to `0.23.0`?

ys:
Or `0.22.1`?

adriancole:
there's an api change to null tracer, so meh I'll defer to others on that

ys:
@jcarres-mdsol hey any ideas on version bumping for this? Kind of want to move forward with this quickly

ys:
Bumped to `0.23.0` for the API change in the null tracer.

ys:
Any news? We kind of want to move forward with this internally and don't want to handle forks and stuff.

jcarres-mdsol:
Sorry, it's good, merging and releasing

Issue #102, opened 2017-03-28
::Trace.record throws "not implemented"


sullerandras:
I try to use the "annotate_plugin" method to add the status code as a binary annotation.
I created a new rails application, added the "zipkin-tracer" gem to gemfile then copied the annotate_plugin setup from the readme:

```ruby
module TestZipkinRuby
  class Application < Rails::Application
    # Settings in config/environments/* take precedence over those specified here.
    # Application configuration should go into files in config/initializers
    # -- all .rb files in that directory are automatically loaded.
    require 'zipkin-tracer'
    zipkin_config = {
      service_name: 'example-app',
      service_port: 443,
      sample_rate: 1,
      json_api_host: 'http://localhost:9411',
      annotate_plugin: lambda do |env, status, response_headers, response_body|
        ep = ::Trace.default_endpoint
        # string annotation
        ::Trace.record(::Trace::BinaryAnnotation.new('http.referrer', env['HTTP_REFERRER'], 'STRING', ep)) # this is line 23
        # integer annotation
        ::Trace.record(::Trace::BinaryAnnotation.new('http.content_size', [env['CONTENT_SIZE']].pack('N'), 'I32', ep))
        ::Trace.record(::Trace::BinaryAnnotation.new('http.status', [status.to_i].pack('n'), 'I16', ep))
      end
    }
    middleware.use ZipkinTracer::RackHandler, zipkin_config
  end
end
```

This throws `not implemented` error at runtime with this backtrace:

```
finagle-thrift (1.4.2) lib/finagle-thrift/tracer.rb:4:in `record'
finagle-thrift (1.4.2) lib/finagle-thrift/trace.rb:43:in `record'
config/application.rb:23:in `block in <class:Application>'
zipkin-tracer (0.21.1) lib/zipkin-tracer/rack/zipkin-tracer.rb:38:in `annotate_plugin'
zipkin-tracer (0.21.1) lib/zipkin-tracer/rack/zipkin-tracer.rb:48:in `trace!'
zipkin-tracer (0.21.1) lib/zipkin-tracer/rack/zipkin-tracer.rb:25:in `block (2 levels) in call'
zipkin-tracer (0.21.1) lib/zipkin-tracer/zipkin_tracer_base.rb:19:in `with_new_span'
zipkin-tracer (0.21.1) lib/zipkin-tracer/rack/zipkin-tracer.rb:24:in `block in call'
[...]

PR #106, opened 2017-05-22
[FIX] pass and use the span in annotate_plugin

elyara:
Addressing the https://github.com/openzipkin/zipkin-ruby/issues/102

Passing the span as @strom suggested


jcarres-mdsol:
Cool!
Can you update the version and changelog so we can release a new gem?

elyara:
@jcarres-mdsol done!

PR #109, opened 2017-06-23
fix memory leak when doing faraday request in a new thread

AlunYou:
A second thought is: if this thread has multiple async http requests, "detect this previous span" logic will break and it will become a mess.
So I submit another solution: detect whether a span has parent, if not, then it needs to be flush immediately. And because of the logic, I have to change some unit tests. But I believe the old application users will not be affected by this logic.

@jfeltesse-mdsol @jcarres-mdsol

### DIFF
--- a/lib/zipkin-tracer/zipkin_tracer_base.rb
+++ b/lib/zipkin-tracer/zipkin_tracer_base.rb
@@ -23,7 +23,16 @@ module Trace
 
     def end_span(span)
       span.close
-      if span.annotations.any?{ |ann| ann.value == Annotation::SERVER_SEND }
+      # if in a thread not handling incoming http requests, it will not have Annotation::SERVER_SEND, so the span
+      # will never be flushed and will cause memory leak.
+      # It will have CLIENT_SEND and CLIENT_RECV if the thread sends out http requests, so use CLIENT_RECV as the sign
+      # to flush the span.
+      has_server_recv_span = spans.any? do |s|
+        s.annotations.any? { |ann| ann.value == Annotation::SERVER_RECV }
+      end
+      if span.annotations.any? { |ann| ann.value == Annotation::SERVER_SEND } ||
+          (!has_server_recv_span &&
+              span.annotations.any? { |ann| ann.value == Annotation::CLIENT_RECV })
         flush!
         reset
       end
--- a/lib/zipkin-tracer/zipkin_tracer_base.rb
+++ b/lib/zipkin-tracer/zipkin_tracer_base.rb
@@ -23,16 +23,12 @@ module Trace
 
     def end_span(span)
       span.close
-      # if in a thread not handling incoming http requests, it will not have Annotation::SERVER_SEND, so the span
+      # If in a thread not handling incoming http requests, it will not have Annotation::SERVER_SEND, so the span
       # will never be flushed and will cause memory leak.
       # It will have CLIENT_SEND and CLIENT_RECV if the thread sends out http requests, so use CLIENT_RECV as the sign
       # to flush the span.
-      has_server_recv_span = spans.any? do |s|
-        s.annotations.any? { |ann| ann.value == Annotation::SERVER_RECV }
-      end
-      if span.annotations.any? { |ann| ann.value == Annotation::SERVER_SEND } ||
-          (!has_server_recv_span &&
-              span.annotations.any? { |ann| ann.value == Annotation::CLIENT_RECV })
+      # If no parent span, then current span needs to flush when it ends.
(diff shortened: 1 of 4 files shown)


lib/zipkin-tracer/zipkin_tracer_base.rb
```diff
@@ -23,7 +23,16 @@ def with_new_span(trace_id, name)
 
     def end_span(span)
       span.close
-      if span.annotations.any?{ |ann| ann.value == Annotation::SERVER_SEND }
+      # if in a thread not handling incoming http requests, it will not have Annotation::SERVER_SEND, so the span
+      # will never be flushed and will cause memory leak.
+      # It will have CLIENT_SEND and CLIENT_RECV if the thread sends out http requests, so use CLIENT_RECV as the sign
```
jcarres-mdsol: Maybe is a cornercase but zipkin allows for local spans. Yo could for instance trace calls to DBs or local programs or anything that is interesting to trace without even calling an external service. I guess this at least solves some uses cases.

Another comment is that spans access a thread local variable which is slow, this change would mean an extra access each span which would make tracing slower for everyone

jcarres-mdsol:
@adriancole @ykitamura-mdsol @jfeltesse-mdsol

lib/zipkin-tracer/zipkin_tracer_base.rb
AlunYou: Thanks, @jcarres-mdsol . I get the point that extra access to "spans" could be slow. How about we move this code into store_span, and save has_server_recv_span in the span? That method will access spans anyway.
```
has_server_recv_span = spans.any? do |s|
        s.annotations.any? { |ann| ann.value == Annotation::SERVER_RECV }
      end
```

I think the memory leak needs to be fixed, because some tools-like gems (like newrelic) can do faraday calls in new thread, and once you use both newrelic and zipkin, you're leaking memory. The user may not realize he's tracing a http call in a gem and he is leaking memory because of that, even if the user only intends to do local tracing.

lib/zipkin-tracer/zipkin_tracer_base.rb
```diff
@@ -23,7 +23,16 @@ def with_new_span(trace_id, name)
...
+      # to flush the span.
+      has_server_recv_span = spans.any? do |s|
+        s.annotations.any? { |ann| ann.value == Annotation::SERVER_RECV }
+      end
+      if span.annotations.any? { |ann| ann.value == Annotation::SERVER_SEND } ||
+          (!has_server_recv_span &&
+              span.annotations.any? { |ann| ann.value == Annotation::CLIENT_RECV })
```
jfeltesse-mdsol: there must be a way to condense all those "any" operations, as it reads now it means we're going to iterate many times over span(s) annotations.
I don't know the typical size of that annotations array but at a glance it looks wasteful.

AlunYou: how about this logic? it doesn't need more iterating on span annotations, just check if there's previous span. For the leaking case, there's only one span, no previous span.

```
def end_span(span)
      span.close
      # if in a thread not handling incoming http requests, it will not have Annotation::SERVER_SEND, so the span
      # will never be flushed and will cause memory leak.
      # It will have CLIENT_SEND and CLIENT_RECV if the thread sends out http requests, so use CLIENT_RECV as the sign
      # to flush the span.
      has_previous_span = spans.size >= 2
      if (has_previous_span && span.annotations.any? { |ann| ann.value == Annotation::SERVER_SEND }) ||
          (!has_previous_span && span.annotations.any? { |ann| ann.value == Annotation::CLIENT_RECV })
        flush!
        reset
      end
    end
```

jcarres-mdsol: LGTM

AlunYou: A second thought is: if this thread has multiple http request, "detect this previous span" logic will break and it will become a mess.
So I submit another solution: detect whether a span has parent, if not, then it needs to be flush immediately.

jcarres-mdsol:
What will happen if  you are a background job that starts all the spans, like the top level of the chain, which for  background jobs may be quite common.
Wont then it span.parent always be nil?

AlunYou:
In this case, I guess these spans should be wrapped by an even top span (which indicates the background job's life time), if you want to report these spans as related. If you do this, these spans will have a parent.
So my idea is : if a span is really top level span, then it has nothing to do with any other spans, so does not have to be reported together with other span, so can be flushed immediately.

PR #112, opened 2017-08-07
Adds tagging of errors to spans which failed status codes

jcarres-mdsol:
See this for background https://github.com/openzipkin/zipkin/pull/1675

@adriancole  maybe other PR provides a better background?
@jfeltesse-mdsol  @ykitamura-mdsol  please review.

I understand there is duplication of code between Faraday and Excon, not only this new tag, others are duplicating also, to be fixed in some other PR.

### DIFF
--- a/lib/zipkin-tracer/excon/zipkin-tracer.rb
+++ b/lib/zipkin-tracer/excon/zipkin-tracer.rb
@@ -29,13 +29,9 @@ module ZipkinTracer
 
     def response_call(datum)
       if span = datum[:span]
-        if status = response_status(datum)
-          span.record_tag(
-            Trace::BinaryAnnotation::STATUS,
-            status,
-            Trace::BinaryAnnotation::Type::STRING,
-            local_endpoint
-          )
+        status = response_status(datum)
+        if status
+          record_response_tags(span, status, local_endpoint)
         end
         span.record(Trace::Annotation::CLIENT_RECV, local_endpoint)
         Trace.tracer.end_span(span)
@@ -47,6 +43,7 @@ module ZipkinTracer
     private
 
     SERVER_ADDRESS_SPECIAL_VALUE = '1'.freeze
+    STATUS_ERROR_REGEXP = /\A(4.*|5.*)\z/.freeze
 
     def b3_headers
       {
@@ -72,6 +69,14 @@ module ZipkinTracer
 
     def response_status(datum)
       datum[:response] && datum[:response][:status] && datum[:response][:status].to_s
+    end
+
+    def record_response_tags(span, status, local_endpoint)
+      span.record_tag(Trace::BinaryAnnotation::STATUS, status, Trace::BinaryAnnotation::Type::STRING, local_endpoint)
+      if STATUS_ERROR_REGEXP.match(status)
+        span.record_tag(Trace::BinaryAnnotation::ERROR, status,
+          Trace::BinaryAnnotation::Type::STRING, local_endpoint)
+      end
     end
 
     def trace!(datum, trace_id)
--- a/lib/zipkin-tracer/faraday/zipkin-tracer.rb
+++ b/lib/zipkin-tracer/faraday/zipkin-tracer.rb
@@ -28,6 +28,8 @@ module ZipkinTracer
     private
 
     SERVER_ADDRESS_SPECIAL_VALUE = '1'.freeze
+    STATUS_ERROR_REGEXP = /\A(4.*|5.*)\z/.freeze
+
 
     def b3_headers
       {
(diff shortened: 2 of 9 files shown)


adriancole:
here's a similar PR https://github.com/openzipkin/brave/pull/398

lib/zipkin-tracer/excon/zipkin-tracer.rb
```diff
@@ -29,14 +29,20 @@ def request_call(datum)
...
+        if status
           span.record_tag(
             Trace::BinaryAnnotation::STATUS,
             status,
             Trace::BinaryAnnotation::Type::STRING,
             local_endpoint
           )
         end
+        if STATUS_ERROR_REGEXP.match(status.to_s)
+          span.record_tag(Trace::BinaryAnnotation::ERROR, "Error #{status}",
```
adriancole: not terribly important, but Error is already the tag name, so not sure prefixing here clarifies. Ex. I just put the status in this case.

lib/zipkin-tracer/excon/zipkin-tracer.rb
```diff
@@ -47,6 +53,7 @@ def response_call(datum)
     private
 
     SERVER_ADDRESS_SPECIAL_VALUE = '1'.freeze
+    STATUS_ERROR_REGEXP = /\A(4.*|5.*)\z/.freeze
```
adriancole: you could also put 1.x here, eventhough it is unlikely.. or reverse it to things that are not 2xx and 3xxx

jcarres-mdsol: Are 1.x errors?

adriancole:
>
> Are 1.x errors?
>
> heh actually you are right. they are not errors. copy-pasted w/o thinking.
I'll fix brave

On Mon, Aug 7, 2017 at 11:07 AM, Jordi Polo Carres <notifications@github.com

jcarres-mdsol:
@adriancole  I've added errors when some exception happen in Faraday. Not sure how to do it in Excon, maybe @ys  can help, could be done in a different PR also.

adriancole:
FWIW, we had this logic for adding "http.status_code" as a tag (only
if it isn't success). This was incorrectly copy/pasted into the error
handling part

if (httpStatus != null && (httpStatus < 200 || httpStatus > 299)) {
  customizer.tag(TraceKeys.HTTP_STATUS_CODE, String.valueOf(httpStatus));
}

jcarres-mdsol:
@adriancole  I would not make 300 errors either, it's a very common non-error code.

adriancole:
@adriancole agree and brave actually doesn't make 300 errors either. I pasted the conditional logic about  "http.status_code" tag which is slightly different than the "error" tag. We currently don't add  "http.status_code" for 2xx, but we do for 3xx. After I fix the bug, error will only be >=400

adriancole:
Actually we might reconsider about 100. 100 is a provisional status (expect
100 continue).. if there's no final status .. it would be an error right?

adriancole:
101 status is a bit more complicated ;) maybe I will open an issue on the
docs repo about 1xx and we can followup later on these codes.

lib/zipkin-tracer/faraday/zipkin-tracer.rb
```diff
@@ -51,12 +53,31 @@ def trace!(env, trace_id)
...
+      msg = "Request timed out."
+      span.record_tag(Trace::BinaryAnnotation::ERROR, msg, Trace::BinaryAnnotation::Type::STRING, local_endpoint)
+      raise e
+    rescue Faraday::ConnectionFailed => e
+      msg = "Request connection failed."
+      span.record_tag(Trace::BinaryAnnotation::ERROR, msg, Trace::BinaryAnnotation::Type::STRING, local_endpoint)
+      raise e
+    rescue Faraday::ClientError => e
+      msg = "Generic Faraday client error."
+      span.record_tag(Trace::BinaryAnnotation::ERROR, msg, Trace::BinaryAnnotation::Type::STRING, local_endpoint)
```
jfeltesse-mdsol: how about making some "record_error" method where you could do

```ruby
rescue Faraday::ClientError => e 
  record_error(span, "Generic Faraday client error.", local_endpoint)
  raise
end
```

also, you shouldn't need to `raise e`, `raise` is enough to re-raise:

```ruby
def boom
  raise "kaboom!"
end

begin
  begin
    boom
  rescue => first_err
    puts "first rescue: #{first_err.inspect}, #{first_err.object_id}"
    raise
  end
rescue => second_err
  puts "second rescue: #{second_err.inspect}, #{second_err.object_id}"
end

# => first rescue: #<RuntimeError: kaboom!>, 70353196231420
# => second rescue: #<RuntimeError: kaboom!>, 70353196231420
```

lib/zipkin-tracer/version.rb
```diff
@@ -1,3 +1,3 @@
 module ZipkinTracer
-  VERSION = '0.25.0'.freeze
+  VERSION = '0.27.0'.freeze
```
jfeltesse-mdsol: last time the version didn't get a bump? 🙈
if we just skip 0.26 it's going to look weird in the gem release history.
maybe we should have a micro-PR bumping the number, release 0.26 and then release 0.27, what do you think?

adriancole:
Sorry for spam but was thinking 404 is in hindsight not a great signal for
error (eventhough most others are). There are a lot of rest apis that use
404 routinely to check if an item exists or not. I think we should exclude
404 from the default codes that result in an error tag. Wdyt?

jcarres-mdsol:
@adriancole how do you use 404 in non-error situations? At least around here will mean it will eventually become a bug in the UI

adriancole:
not found is routinely used to say a resource doesn't exist. This may not
be an error because for example think of a resource which you are checking
to see if an entry exists or not. For example zipkin's trace api. You query
and see if the trace exists or not. This doesn't imply an error, just that
it doesn't exist.

jcarres-mdsol:
Umm, it's unlikely I'll go to zipkin and add a random trace_id. It's more likely there is some UI which links to Zipkin and that UI should have not display that trace_id to me if that is a dead link which can not retrieve data.
Also Zipkin is an odd case because of the sampling rate

adriancole:
zipkin api is just one example. Here's another. cloud storage. have you
ever requested something to see if exists in storage and if not done
something differently?

jcarres-mdsol:
No really. I would not retrieve an object if I have no clue if it's there. I'd list objects or search an object list. So when I try to retrieve, I really believe is there and not being there (they may have just deleted it in the milliseconds between me checking the list and retrieving the data) that would mean an error to me.

jcarres-mdsol:
Maybe should add a configuration key for this, but it's kind of overblowing it :(

adriancole:
:) ok anyway we can wait until someone complains. I used to run a cloud
storage library and have different experiences on this matter.

adriancole:
fwiw consider a bucket with N pages of objects. it isn't efficient to
use this approach to test if a resource exists or not, and not all
scenarios for "not exist" are race conditions that can't resolve (via
conditional put). you can have a look at the ruby fog library which is
similar to jclouds which I used to work on a lot. That said, yeah it
is reasonable to not optimize for this kind of rest api until people
complain.

jcarres-mdsol:
What about 1xx. Add them as errors or no? I do not have opinion either way, I've hardly encounter any in my life

adriancole:
>
> What about 1xx. Add them as errors or no? I do not have opinion either
> way, I've hardly encounter any in my life
>
At least one of them isn't an error (switch protocols is for web sockets
iirc), so I'd leave 1xx out of error range for now. I'll open an issue on
the docs side to look further into it (just as reference)

adriancole:
PS I will leave brave also reporting all 4xx as errors until/unless we
hear back otherwise. Will let you know if that happens. (that said I
will take 1xx out of error range for now)

jcarres-mdsol:
So good to merge?

adriancole:
please do! thanks for the help, too. I'm later going to announce the error counting version of zipkin and will mention this is supported in latest zipkin-ruby

jcarres-mdsol:
Cool 👯

ykitamura-mdsol:
looks good to me 🐨

ys:
Thanks @jcarres-mdsol!

I was on vacation, sorry for the late response but it looks good!

PR #114, opened 2017-08-16
Add basic authentication support for json tracer

jeroennijhof:
This pull request will add basic authentication support to the json tracer so you are still able to use zipkin when the server is protected with basic auth.

### DIFF
--- a/lib/zipkin-tracer/config.rb
+++ b/lib/zipkin-tracer/config.rb
@@ -6,9 +6,9 @@ module ZipkinTracer
   # Configuration of this gem. It reads the configuration and provides default values
   class Config
     attr_reader :service_name, :service_port, :json_api_host,
-      :zookeeper, :sample_rate, :logger, :log_tracing,
-      :annotate_plugin, :filter_plugin, :whitelist_plugin,
-      :sampled_as_boolean, :record_on_server_receive,
+      :json_api_user, :json_api_password, :zookeeper, :sample_rate,
+      :logger, :log_tracing, :annotate_plugin, :filter_plugin,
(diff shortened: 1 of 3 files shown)


adriancole:
I like the idea here, but zipkin's server doesn't use basic auth (it hints at a custom impl). This could confuse people as there's no out-of-box service that supports this.

Even if a server implementation was protected, it might not use basic auth to do so (ex it could be oauth or api token).

Maybe give some background around your setup? Is there a more generic way to accomplish this? For example, in java we allow users to provide an interceptor which they can choose to use for basic auth if they have a custom setup. I'd like to see a way that helps you, but that doesn't confuse people via keys exposed for setups they can't run OOB.

There's more background about this here: https://github.com/openzipkin/zipkin/issues/782

jcarres-mdsol:
Also this code seems to force everyone to use basic auth, as pointed out by Adrian, that's the very least not a common setup.

jeroennijhof:
@jcarres-mdsol nobody is forcing anything, it's optional.

@adriancole we have a zipkin server running behind a reverse proxy which has basic auth enabled as well as ssl. I was sort of surprised zipkin does not have any sort of authentication. We need at least some sort of authentication since the data may not be visible to some of our customers. Any better idea's?

adriancole:
I had mentioned an interceptor approach. For example a function that can
decorate the http request with an auth token.

Bear in mind http is not the only transport even if simplest. Ex sqs is a
valid span transport and has auth as do others.

jeroennijhof:
@adriancole thanks, I will go for the interceptor approach.

adriancole:
FWIW you are right. Others are using proxies and interceptor approach for
http auth, I think at least @devinsba

PR #119, opened 2018-02-11
Stringify trace values

ykitamura-mdsol:
## Background
Sending numeric trace values causes errors in Zipkin but the errors are not logged:
```ruby
span.record_tag 'id', 123
```

## Update summary
- Converting trace values to string in `Trace::Span#record` and `Trace::Span#record_tag` methods.
- ~~Add logging when Zipkin returns 400 Bad Request to catch `Cannot decode spans due to IllegalArgumentException` errors~~

@adriancole @jcarres-mdsol  @jfeltesse-mdsol @ssteeg-mdsol

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 0.27.2
+* Convert trace values to string
+* Add logging on 400 Bad Request
+
 # 0.27.1
 * Rescue connection errors when sending information to Zipkin fails
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,5 @@
 # 0.27.2
 * Convert trace values to string
-* Add logging on 400 Bad Request
 
 # 0.27.1
 * Rescue connection errors when sending information to Zipkin fails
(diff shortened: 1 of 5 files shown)


jfeltesse-mdsol:
Same, how often do 400 happen there?
Do we log failures at other places the same way?

jcarres-mdsol:
I'm worried if some day zipkin server returns a huge body on that 400 and we log some horrendous blob

adriancole:
it would be most likely that the blob on 400 would be the result of a large
blob sent :) one thing you can do is character limit it to XXX amount
and/or make sure the message is a fine level in the log

On Tue, Feb 13, 2018 at 9:43 AM, Jordi Polo Carres <notifications@github.com

ykitamura-mdsol:
I think adding `to_s` will reduce the majority of 400 errors which are raised on processing malformed spans:
https://github.com/openzipkin/zipkin/blob/release-2.4.6/zipkin/src/main/java/zipkin/internal/Collector.java#L113 

I'm going to remove the logging. 
I might come back if we start getting many 400 errors again.

PR #121, opened 2018-05-10
Support 128-bit trace IDs

ykitamura-mdsol:
Add the `:trace_id_128bit` configuration option to support 128-bit trace IDs

https://github.com/openzipkin/b3-propagation/issues/6

@adriancole @jcarres-mdsol @ssteeg-mdsol @cabbott


jcarres-mdsol:
Beautiful, I'll let @adriancole take a look, else will merge in a couple of days

Issue #96, opened 2017-01-06
How to set component_span with Rack middleware?


georgeu2000:
We have the following config for a Sinatra App:

```
zipkin_config = { service_name:ENV[ '...' ],
                  service_port:settings.port,
                  sampled_as_boolean:false,
                  sample_rate: 1,
                  json_api_host: ENV[ '..' ]}
use ZipkinTracer::RackHandler, zipkin_config
```

The span is displayed as "post", which I guess comes from the request method. How can I set it to request method and request path?

adriancole:
the span name is currently hard-set. It could be possible to add a naming function instead, which could compose the value. A lot of instrumentation do things like this.

https://github.com/openzipkin/zipkin-ruby/blob/master/lib/zipkin-tracer/rack/zipkin-tracer.rb#L24

georgeu2000:
What if we change it to

    @tracer.with_new_span(trace_id,"#{env['REQUEST_METHOD']} #{env['PATH_INFO']}".downcase) do ...

Since this is a default, it might be better to more explicit...? (Although I have zero context so this is just for my use case.)

I would be happy to do a PR.

adriancole:
PATH_INFO is likely to be variable, right? If so, it would explode the
cardinality of span names which would cause usability (and also UI
performance problems). The span name is a label in the UI.

I think the default should be left alone, but allow a function of env to be
passed to customize. For example, that could clean up or otherwise format
PATH_INFO such that it doesn't repeat endlessly (such as something with an
identifier in it would). For those who happen to have low cardinality
paths, they could choose to put the function you mention.

sg?

adriancole:
seems there's a long discussion of how to get what's currently referred to as "http.route" from the rails env https://stackoverflow.com/questions/42691729/how-to-find-current-abstract-route-in-rails-middware

see https://github.com/openzipkin/zipkin-api/pull/42 for http.route

jmhon08:
I'm not exactly clear how to use the change in https://github.com/openzipkin/zipkin-api/pull/42/files to now get the route in the span name. Would it be a simple change in zipkin-ruby or very involved like the one for brave (https://github.com/openzipkin/brave/pull/602/files)

adriancole:
here's the route-based span name logic we use in java if that helps? https://github.com/openzipkin/brave/blob/master/instrumentation/http/src/main/java/brave/http/HttpParser.java#L94

jmhon08:
[...]

PR #122, opened 2018-06-20
Add abstract route to span name https://github.com/openzipkin/zipkin-ruby/issues/96

jmhon08:
screen shot 2018-06-20 at 10 19 41 am

### DIFF
--- a/lib/zipkin-tracer/rack/zipkin-tracer.rb
+++ b/lib/zipkin-tracer/rack/zipkin-tracer.rb
@@ -30,7 +30,7 @@ module ZipkinTracer
         if !trace_id.sampled? || !routable_request?(env)
           @app.call(env)
         else
-          @tracer.with_new_span(trace_id, env[REQUEST_METHOD].to_s.downcase) do |span|
+          @tracer.with_new_span(trace_id, "#{env[REQUEST_METHOD].to_s.downcase} #{route(env)}".strip) do |span|
             trace!(span, zipkin_env) { @app.call(env) }
           end
         end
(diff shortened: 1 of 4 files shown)


jmhon08:
@ykitamura-mdsol @jcarres-mdsol could you please take a look?

lib/zipkin-tracer/rack/zipkin-tracer.rb
```diff
@@ -30,7 +30,7 @@ def call(env)
         if !trace_id.sampled? || !routable_request?(env)
           @app.call(env)
         else
-          @tracer.with_new_span(trace_id, env[REQUEST_METHOD].to_s.downcase) do |span|
+          @tracer.with_new_span(trace_id, "#{env[REQUEST_METHOD].to_s.downcase} #{route(env)}".strip) do |span|
```
adriancole: maybe put in a unit test for this, as we don't want a span name of "get nil" or "get " If there's no route, the method alone should be the name.

spec/lib/rack/zipkin-tracer_spec.rb
```diff
@@ -75,30 +79,48 @@ def expect_host(host)
   context 'Using Rails' do
     subject { middleware(app) }
 
-    context 'accessing a valid URL of our service' do
+    context 'accessing a valid URL "/" of our service' do
       before do
-        allow(middleware(app)).to receive(:routable_request?).and_return(true)
```
jmhon08: Previously, if I changed this to and_return(false), I would still see routable_request?(env) return true in rack/zipkin-tracer.rb

lib/zipkin-tracer/rack/zipkin-tracer.rb
jmhon08: Done in latest commit, can you please review?

jmhon08:
@adriancole @ykitamura-mdsol @jcarres-mdsol could someone pls merge this?

adriancole:
thanks! while not a rails expert, the change you made wrt route looks sound. folks can feel free to give more feedback post merge.

Issue #128, opened 2018-10-23
Feature request: Grape compatibility


yellowred:
Hi,

This gem helps to trace incoming requests, however in case of API made with popular API framework Grape it fails to help. The reason for that is Grape having it's own storage for routes.

Before release `0.29.0` of zipkin-ruby it was possible to patch `routable_request?` method adding:
```
return true if Grape::API.recognize_path(path_info)
```
But with the new version `get_route` also needs to be patched to get rid of extra slash in the beginning of `request.path` (env[SCRIPT_NAME] becomes updated). It seems either Grape is included in the roadmap or it will become incompatible with zipkin.

adriancole:
is this something you can help with?

PR #129, opened 2018-10-23
Create a cleaner temporary request object

yellowred:
Using `Rails.application.routes.router.recognize` is not safe as it updates the passed argument, which is request in our case. Specifically it updates `rails_req.script_name` from `""` to `"/"` which makes `request.path` to have double slash in the beginning. To solve this issue it has been created a temporary clean request object which will be disposed later.

See also: https://github.com/openzipkin/zipkin-ruby/pull/122/files (PR that has the same line before).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 0.29.1
+* Patch abstract route to work with Grape
+
 # 0.29.0
 * Add abstract route to span name
 * Fix not to raise "NoMethodError" for Rails:Module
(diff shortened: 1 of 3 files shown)


yellowred:
Related to this issue: 
https://github.com/openzipkin/zipkin-ruby/issues/128

jcarres-mdsol:
@yellowred  fix looks good, can you make it something like 
```
env = {
"PATH_INFO" => env[ZipkinTracer::RackHandler::PATH_INFO], 
"REQUEST_METHOD" => env[ZipkinTracer::RackHandler::REQUEST_METHOD]
}
```
so we don't have long lines? Github not happy about that

jcarres-mdsol:
Also, please update version and changelog

jcarres-mdsol:
Thanks!

PR #130, opened 2018-10-25
Bug revise of monkey patch without Rails gem

aiharak:
such as rails-html-sanitizer gem uses Rails constant without rails gem
ref. https://github.com/rails/rails-html-sanitizer/blob/master/lib/rails/html/sanitizer.rb#L1


jcarres-mdsol:
Good.
I think you need to:
- Create something like:
```
def running_inside_rails?
   defined?(Rails) && defined?(Rails.application)
end
```
- Use that method in the two lines you have modified
- Pass the tests
- Update minor version and changelog

jcarres-mdsol:
I'm closing this, we are going to revisit this section of the code soon

PR #131, opened 2018-12-06
Add http.method to client annotations

ykitamura-mdsol:
Since abstract route is added to span names in https://github.com/openzipkin/zipkin-ruby/pull/122, adding the [`HTTP method`](https://zipkin.io/public/thrift/v1/zipkinCore.html#Const_HTTP_METHOD) to client annotations.

@adriancole @jcarres-mdsol

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,9 @@ language: ruby
 jdk:
   - oraclejdk8
 rvm:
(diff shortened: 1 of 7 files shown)


jcarres-mdsol:
nice, really useful information

adriancole:
good

PR #132, opened 2018-12-27
Add example for ruby

luong-komorebi:
Because there is no example in the openzipkin org. 
https://github.com/openzipkin?utf8=%E2%9C%93&q=example&type=&language=ruby will return no result

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,5 @@
 Gemfile.lock
 
 .byebug_history
+
+example/dump.rdb
--- a/example/Gemfile
+++ b/example/Gemfile
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+source "https://rubygems.org"
+
+git_source(:github) {|repo_name| "https://github.com/#{repo_name}" }
+
+# gem "rails"
+
+gem "sinatra", "~> 2.0"
+
+gem "thin", "~> 1.7"
+
+gem "zipkin-tracer", "~> 0.30.0"
+
+gem "sidekiq", "~> 5.2"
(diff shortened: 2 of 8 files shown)


adriancole:
are you up to making a port of our standard example? if so, I could open a repo for you. In that repo, you could make a branch that also uses sidekiq.

luongvo209:
@adriancole yeah I am pleased to do that. But before you open that repo, I suggest that I would open a repo on my side (that would be luongvo209/zipkin-ruby-example). I would post a reply here and tag you into that repo when everything is finished and ready to be reviewed. Then after everything looks good to go, we would migrate it to the openzipkin org. How does that sound to you?

adriancole:
sounds good to me

luongvo209:
@jcarres-mdsol Yes definitely. 
Also, if you have plans to write an example, go ahead. Follow your plan. Don't mind me 👍

jcarres-mdsol:
No plan, waiting for your example 👯‍♂️

adriancole:
I created the canonical example here. feel free to help make things more idiomatic, though the routes as they are defined is important (as it lets other people help troubleshoot tracer issues even if they don't know ruby) https://github.com/openzipkin/zipkin-ruby-example

luong-komorebi:
Sorry for letting this hang for too long. I was too busy with OpenCensus. I think we should work more on your example and allow me to close this so efforts are focused in one place only.

adriancole:
thanks for the spike, Luong and no worries.. I needed the example to verify
a bug so scrapped it together. Your code here helped.

PS one approach that works really well is to create a branch for extra
integrations. This provides a "diff" approach without needing to clutter
the main example which should both be canonical and simple as possible.
Here's an example.. you could imagine adding a diff that adds sidekiq
https://github.com/openzipkin/sleuth-webmvc-example

PR #134, opened 2019-01-08
Use proper license Apache-2.0 instead of Apache

jcarres-mdsol:
Use the proper license name from https://spdx.org/licenses/
The LICENSE file is already Apache 2.0 , this is describing what's already a reality, not a change in licensing.

@adriancole  please merge when satisfied.


adriancole:
thanks a lot doing the chores!

PR #138, opened 2019-02-26
Fixes JSON v1 encoding of remote address binary annotations

adriancole:
The only use case for BOOL binary annotations is identifying the remote
endpoint. In thrift encoding, BOOL true is represented by 1. This
accidentally carried into JSON encoding, which tripped up endpoint
interpretation. This fixes it by special casing BOOL.

See https://github.com/openzipkin/zipkin/issues/2414

### DIFF
--- a/lib/zipkin-tracer/trace.rb
+++ b/lib/zipkin-tracer/trace.rb
@@ -94,7 +94,8 @@ module Trace
     def to_h
       {
         key: @key,
-        value: @value,
+        # special case address binary annotations which use BOOL type
+        value: @annotation_type == Trace::BinaryAnnotation::Type::BOOL ? true : @value,
         endpoint: host.to_h
       }
     end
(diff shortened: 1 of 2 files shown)


adriancole:
I manually tested by changing the gem version up a notch, installing, running zipkin-ruby-example, and noticing it looking fine

jcarres-mdsol:
Interesting approach, as we are hardcoding the `1` value, I'd rather change that. I'll do that in a couple of days. Leaving this open till next pr is available.

adriancole:
> Interesting approach, as we are hardcoding the 1 value, I'd rather change that. I'll do that in a couple of days. Leaving this open till next pr is available.

okie dokie, though meanwhile anyone using this will be broke

jcarres-mdsol:
I'm closing this in favor of #139 But more of your ruby is always welcome!

adriancole:
:)

Issue #137, opened 2019-02-26
why is :service_port REQUIRED?


adriancole:
In the data model, port is not required. Is there a reason why we bother people about this? or is it just a historical effect?

jcarres-mdsol:
Fixed by #139

PR #139, opened 2019-02-27
 Fix 'sa' annotation encoding and remove the ':service_port' configuration

ykitamura-mdsol:
Changed the hardcoded value for 'sa' annotation to `true`:
https://github.com/openzipkin/zipkin/issues/2414

~~Updated the ':service_port' configuration to be optional~~  Removed the ':service_port' configuration: https://github.com/openzipkin/zipkin-ruby/issues/137

@adriancole @jcarres-mdsol

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ use ZipkinTracer::RackHandler, config # config is optional
 where `Rails.config.zipkin_tracer` or `config` is a hash that can contain the following keys:
 
 * `:service_name` **REQUIRED** - the name of the service being traced. There are two ways to configure this value. Either write the service name in the config file or set the "DOMAIN" environment variable (e.g. 'test-service.example.com' or 'test-service'). The environment variable takes precedence over the config file value.
-* `:service_port` **REQUIRED** - the port of the service being traced (e.g. 80 or 443)
+* `:service_port` - the port of the service being traced (e.g. 80 or 443)
 * `:sample_rate` (default: 0.1) - the ratio of requests to sample, from 0 to 1
 * `:json_api_host` - hostname with protocol of a zipkin api instance (e.g. `https://zipkin.example.com`) to use the JSON tracer
 * `:zookeeper` - the address of the zookeeper server to use by the Kafka tracer
--- a/README.md
+++ b/README.md
@@ -18,7 +18,6 @@ use ZipkinTracer::RackHandler, config # config is optional
 where `Rails.config.zipkin_tracer` or `config` is a hash that can contain the following keys:
 
 * `:service_name` **REQUIRED** - the name of the service being traced. There are two ways to configure this value. Either write the service name in the config file or set the "DOMAIN" environment variable (e.g. 'test-service.example.com' or 'test-service'). The environment variable takes precedence over the config file value.
-* `:service_port` - the port of the service being traced (e.g. 80 or 443)
 * `:sample_rate` (default: 0.1) - the ratio of requests to sample, from 0 to 1
 * `:json_api_host` - hostname with protocol of a zipkin api instance (e.g. `https://zipkin.example.com`) to use the JSON tracer
 * `:zookeeper` - the address of the zookeeper server to use by the Kafka tracer
@@ -64,7 +63,6 @@ Sidekiq tracing can be turned on by adding ZipkinTracer::Sidekiq::Middleware to 
 ```ruby
 zipkin_tracer_config = {
   service_name: 'service',
-  service_port: 3000,
   json_api_host: 'http://zipkin.io',
   traceable_workers: [:MyWorker, :MyWorker2],
   sample_rate: 0.5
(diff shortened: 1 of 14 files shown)


lib/zipkin-tracer/faraday/zipkin-tracer.rb
```diff
@@ -25,7 +25,7 @@ def call(env)
 
     private
 
-    SERVER_ADDRESS_SPECIAL_VALUE = '1'.freeze
+    SERVER_ADDRESS_SPECIAL_VALUE = true
```
adriancole: I assume thrift will encode this properly, but curious.. did you check?

adriancole: I think somewhere else we do a .to_s on the input, which probably will result in this not working as expected. might be wrong. This is why in #138 I dodged in a different way

ykitamura-mdsol: Thanks! I'll fix not to do `.to_s` on booleans.

adriancole:
thanks, we still need to check to make sure thrift encodes that boolean properly. because I am not that familiar with infrastructure, it took me a lot of manual work to verify the other pull request. Maybe you know a quick way to test this to ensure the remote endpoint ends up ok (ex by looking at result of https://github.com/openzipkin/zipkin-ruby-example with this patched version, but using thrift to send not json)

adriancole:
one way to test is to make a test before this one that compares the result of serializing a span with "sa" to thrift vs a hex encoded value. Then, just make sure that still passes after this one.

jcarres-mdsol:
We deleted the thrift dependency some versions ago and noone has complained so I do not thing anyone was using it.
We are talking about Ruby here, performance is never the first consideration, I'd be shocked when someone opens an issue because they use thrift + ruby .

adriancole:
> We deleted the thrift dependency some versions ago and noone has
> complained so I do not thing anyone was using it.
>
I thought kafka still uses thrift encoding.. is that not the case? If it is
by accident, we should switch it to json..

adriancole:
here looks like it still uses thrift https://github.com/openzipkin/zipkin-ruby/blob/master/lib/zipkin-tracer/zipkin_kafka_tracer.rb#L37

if we switch this to a json list all good..

jcarres-mdsol:
Oh, forgot about these guys. Yes, if the thrift library is added to the project which also contains zipkin-ruby library, then this adaptor loads it dynamically and uses it. So I guess we will need to support it or change this. It is scary because I've not kafka around, difficult to adhoc this.

adriancole:
when I test kafka I use this...
https://github.com/openzipkin/docker-zipkin#kafka

ykitamura-mdsol:
got this error when requiring 'finagle-thrift':
```
Failure/Error: require 'finagle-thrift'

TypeError:
  superclass mismatch for class NullTracer
```
seems we lose the compatibility with thrift now.

adriancole:
maybe different issue but converting Kafka to send same payload as http
probably good idea. then no games

jcarres-mdsol:
I think so, let's kill Thrift definitely, none of the maintainers have a decent setup to test it out and noone has stepped up to fill the role, I don't think it will be missed but if so, I'd gladly merge any PR related to it.

ykitamura-mdsol:
it sounds like a separate story. I don't have a JRuby project, cannot test properly.

lib/zipkin-tracer/config.rb
```diff
@@ -16,7 +16,7 @@ def initialize(app, config_hash)
       # The name of the current service
       @service_name      = config[:service_name]
       # The port where the current service is running
-      @service_port      = config[:service_port]      || DEFAULTS[:service_port]
+      @service_port      = config[:service_port]
```
jcarres-mdsol: I dont thing we need this at all. This is only there for local and we are not sending it now, right?

README.md
```diff
@@ -18,7 +18,7 @@ use ZipkinTracer::RackHandler, config # config is optional
 where `Rails.config.zipkin_tracer` or `config` is a hash that can contain the following keys:
 
 * `:service_name` **REQUIRED** - the name of the service being traced. There are two ways to configure this value. Either write the service name in the config file or set the "DOMAIN" environment variable (e.g. 'test-service.example.com' or 'test-service'). The environment variable takes precedence over the config file value.
-* `:service_port` **REQUIRED** - the port of the service being traced (e.g. 80 or 443)
+* `:service_port` - the port of the service being traced (e.g. 80 or 443)
```
jcarres-mdsol: can't we just delete this?

jcarres-mdsol:
Yes, let's finish this one and make the migration in some other story

README.md
ykitamura-mdsol: updated in https://github.com/openzipkin/zipkin-ruby/pull/139/commits/29af492d4f5a0f134b7799aae196a5b8b8ba961c

PR #141, opened 2019-03-07
Switch to Zipkin v2 span format

ykitamura-mdsol:
Use the simplified Zipkin v2 span format to send traces to Zipkin server.

See https://github.com/openzipkin/zipkin/issues/1499 for details

### DIFF
--- a/lib/zipkin-tracer/hostname_resolver.rb
+++ b/lib/zipkin-tracer/hostname_resolver.rb
@@ -1,17 +1,17 @@
 require 'resolv'
 
 module ZipkinTracer
-  # Resolves hostnames in the endpoints of the annotations.
+  # Resolves hostnames in the endpoints of the spans.
   # Resolving hostnames is a very expensive operation. We want to store them raw in the main thread
   # and resolve them in a different thread where we do not affect execution times.
   class HostnameResolver
     def spans_with_ips(spans)
       host_to_ip = hosts_to_ipv4(spans)
 
-      each_annotation(spans) do |annotation|
-        hostname = annotation.host.ipv4
+      each_endpoint(spans) do |endpoint|
+        hostname = endpoint.ipv4
         unless resolved_ip_address?(hostname.to_s)
-          annotation.host.ipv4 = host_to_ip[hostname]
+          endpoint.ipv4 = host_to_ip[hostname]
         end
       end
     end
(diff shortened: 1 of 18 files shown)


lib/zipkin-tracer/hostname_resolver.rb
```diff
@@ -28,24 +28,19 @@ def resolved_ip_address?(ip_string)
         ip_string.to_i.to_s == ip_string
     end
 
-    def each_annotation(spans, &block)
+    def each_endpoint(spans, &block)
```
adriancole: maybe add a comment about what this is doing? I'm guessing it is lazy parsing?

lib/zipkin-tracer/rack/zipkin-tracer.rb
```diff
@@ -55,12 +55,11 @@ def annotate_plugin(span, env, status, response_headers, response_body)
 
     def trace!(span, zipkin_env, &block)
       trace_request_information(span, zipkin_env)
-      span.record(Trace::Annotation::SERVER_RECV)
+      span.kind = Trace::Span::Kind::SERVER
```
adriancole: is there a way to tell if the incoming span ID was shared (from b3 headers)? if so, add span.shared=true

lib/zipkin-tracer/trace.rb
```diff
@@ -199,12 +162,29 @@ def to_i; @i128; end
 
   # A span may contain many annotations
   class Span
-    attr_accessor :name, :annotations, :binary_annotations, :debug
+    module Tag
+      METHOD = "http.method".freeze
+      PATH = "http.path".freeze
+      STATUS = "http.status".freeze
+      LOCAL_COMPONENT = "lc".freeze
```
adriancole: ls is no longer needed unless there's a value for it. sometimes we had "lc"->"" just to add an endpoint to a span.

lib/zipkin-tracer/trace.rb
```diff
@@ -199,12 +162,29 @@ def to_i; @i128; end
 
   # A span may contain many annotations
   class Span
-    attr_accessor :name, :annotations, :binary_annotations, :debug
+    module Tag
+      METHOD = "http.method".freeze
+      PATH = "http.path".freeze
+      STATUS = "http.status".freeze
```
adriancole: this is wrong, it should be http.status_code though possibly was always incorrect here

lib/zipkin-tracer/hostname_resolver.rb
ykitamura-mdsol: I think this comment is explaining:
https://github.com/openzipkin/zipkin-ruby/blob/f0435f983ca63644db0fdf911f5bb52b2b9df87b/lib/zipkin-tracer/hostname_resolver.rb#L5-L6

lib/zipkin-tracer/rack/zipkin-tracer.rb
ykitamura-mdsol: updated in https://github.com/openzipkin/zipkin-ruby/pull/141/commits/1139cc55956a570a48dca0230f25ac64ef2b7b2a

lib/zipkin-tracer/trace.rb
ykitamura-mdsol: It seems we are still using it in our library. Added TODO in https://github.com/openzipkin/zipkin-ruby/pull/141/commits/2cd021ef2861135539effbcd8d9088c48d308a81

lib/zipkin-tracer/trace.rb
ykitamura-mdsol: updated https://github.com/openzipkin/zipkin-ruby/pull/141/commits/d4b62540cef69d5c012ee05a2355e090339c1f56

ykitamura-mdsol:
It seems we always have `duration` in spans:
https://github.com/openzipkin/zipkin-ruby/blob/76ea4940d019f03d32ec036f6f6a4e114bc79669/lib/zipkin-tracer/trace.rb#L213-L215

jcarres-mdsol:
I am going to merge with the `lc` TODO. Moving to V2 is a huge step forward

adriancole:
agreed. this is a huge help

PR #142, opened 2019-03-20
Add the ':additional_paths' configuration option to enable tracing on non-Rails-routable paths

ykitamura-mdsol:
Add the `:additional_paths` configuration option to enable tracing on non-Rails-routable paths, such as in apps using [Roda](https://github.com/jeremyevans/roda), [Sinatra](https://github.com/sinatra/sinatra).

This is another approach to solve the issue in https://github.com/openzipkin/zipkin-ruby/pull/130

### DIFF
--- a/lib/zipkin-tracer/config.rb
+++ b/lib/zipkin-tracer/config.rb
@@ -9,7 +9,7 @@ module ZipkinTracer
       :zookeeper, :sample_rate, :logger, :log_tracing,
       :annotate_plugin, :filter_plugin, :whitelist_plugin,
       :sampled_as_boolean, :record_on_server_receive,
-      :kafka_producer, :kafka_topic, :trace_id_128bit
+      :kafka_producer, :kafka_topic, :trace_id_128bit, :additional_paths
 
     def initialize(app, config_hash)
       config = config_hash || Application.config(app)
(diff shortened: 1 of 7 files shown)


lib/zipkin-tracer/config.rb
```diff
@@ -26,8 +26,6 @@ def initialize(app, config_hash)
       @sample_rate       = config[:sample_rate]       || DEFAULTS[:sample_rate]
       # A block of code which can be called to do extra annotations of traces
       @annotate_plugin   = config[:annotate_plugin]   # call for trace annotation
-      @filter_plugin     = config[:filter_plugin]     # skip tracing if returns false
-      @whitelist_plugin  = config[:whitelist_plugin]  # force sampling if returns true
```
ykitamura-mdsol: no need to set twice (see below) 😎

jfeltesse-mdsol:
The naming might be too broad to explain what this does but I can't find a better alternative.

Otherwise LGTM.

jcarres-mdsol:
It is not bad, I'm concerned if some day we want to base this decision on other thing other than paths, like headers or something, then we need to change this somehow...

jcarres-mdsol:
I've read the related code, and it is a mess. Probably my fault.

There are different places where we do different things. We should consolidate.

The `routable?` information from the rack middleware should be moved here https://github.com/openzipkin/zipkin-ruby/blob/master/lib/zipkin-tracer/rack/zipkin_env.rb

And now we have all the stuff in there.

Then here this would be the logic:
trace.sampled = parent.sampled && !filter_plugin() && (whitelist_plugin() || routable? )

This is set to the trace once here and then we can delete from faraday and excon the check on if something is routable.

We thus do not need to add a new configuration. We need to fix how the whitelist plugin is used.

To achieve the original intent, someone will pass something like 
->(env) { env["PATH_INFO"] ~ whitelisted_regex }

to the whitelist plugin

ykitamura-mdsol:
@jcarres-mdsol `whitelist_plugin` is used for force sampling now but do you mean you are going to make a backward-incompatible change?
And I don't see any routable checks in faraday or excon...
Would you please make a PR for this?

jcarres-mdsol:
True. 
I'd just move routable_requests? to zipkin_env and change https://github.com/openzipkin/zipkin-ruby/blob/master/lib/zipkin-tracer/rack/zipkin_env.rb#L67 to this 
`new_sampled_header_value(force_sample? || current_trace_sampled? && !filtered? && routable_request?)`

ykitamura-mdsol:
closing in favor of #143

PR #143, opened 2019-03-26
Whitelist would have precedence to routing

jcarres-mdsol:
Alternative implementation to #142 

@jfeltesse-mdsol @ssteeg-mdsol @piao-mdsol @adriancole @ykitamura-mdsol

### DIFF
--- a/lib/zipkin-tracer/application.rb
+++ b/lib/zipkin-tracer/application.rb
@@ -1,17 +1,18 @@
 module ZipkinTracer
-
   # Useful methods on the Application we are instrumenting
   class Application
-    # If the request is not valid for this service, we do not what to trace it.
-    def self.routable_request?(path_info, http_method)
+    # Does our framework know if this request will be routed to a controller?
+    def self.routable_request?(env)
       return true unless defined?(Rails) # If not running on a Rails app, we can't verify if it is invalid
+      path_info = env[ZipkinTracer::RackHandler::PATH_INFO]
+      http_method = env[ZipkinTracer::RackHandler::REQUEST_METHOD]
       Rails.application.routes.recognize_path(path_info, method: http_method)
       true
     rescue ActionController::RoutingError
       false
     end
 
-    def self.get_route(env)
+    def self.route(env)
       return nil unless defined?(Rails)
       stub_env = {
         "PATH_INFO" => env[ZipkinTracer::RackHandler::PATH_INFO],
--- a/lib/zipkin-tracer/application.rb
+++ b/lib/zipkin-tracer/application.rb
@@ -1,7 +1,7 @@
 module ZipkinTracer
   # Useful methods on the Application we are instrumenting
   class Application
-    # Does our framework know if this request will be routed to a controller?
+    # Determines if our framework knows whether the request will be routed to a controller
     def self.routable_request?(env)
       return true unless defined?(Rails) # If not running on a Rails app, we can't verify if it is invalid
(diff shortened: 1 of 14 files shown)


lib/zipkin-tracer/excon/zipkin-tracer.rb
```diff
@@ -37,7 +37,6 @@ def response_call(datum)
 
     private
 
-    SERVER_ADDRESS_SPECIAL_VALUE = true
```
jcarres-mdsol: I think #139 got rid of these

lib/zipkin-tracer/rack/zipkin-tracer.rb
```diff
@@ -56,7 +48,6 @@ def annotate_plugin(span, env, status, response_headers, response_body)
     def trace!(span, zipkin_env, &block)
       trace_request_information(span, zipkin_env)
       span.kind = Trace::Span::Kind::SERVER
-      span.record('whitelisted') if zipkin_env.force_sample?
```
jcarres-mdsol: This could be potentially breaking change but also potentially misleading, so taking it out, there is an annotation plugin which could bring this again if needed.

lib/zipkin-tracer/rack/zipkin_env.rb
```diff
@@ -64,12 +60,21 @@ def sampled_header_value(parent_trace_sampled)
       if parent_trace_sampled  # A service upstream decided this goes in all the way
         parent_trace_sampled
       else
-        new_sampled_header_value(force_sample? || current_trace_sampled? && !filtered?)
+        new_sampled_header_value(force_sample? || current_trace_sampled? && !filtered? && routable_request?)
```
ykitamura-mdsol: > Whitelist plugin ensure a request is traced even if it is not routable

If so, I think this line should be:
```ruby
new_sampled_header_value(current_trace_sampled? && !filtered? && (routable_request? || force_sample?))
```

and you need to update README.md and comments 😄

jcarres-mdsol: It depends on what we want whitelist. to do, if its only to overwrite the routing or to overwrite any decision including sampling. I think it makes more sense as the later, something like when your request hits `/procesingMoney` then it always samples 100%

ssteeg-mdsol: Is the ordering of `current_trace_sampled? && !filtered? && routable_request?` performance related or could it can be changed to `current_trace_sampled? && routable_request? && !filtered?`

lib/zipkin-tracer/application.rb
```diff
@@ -1,17 +1,18 @@
 module ZipkinTracer
-
   # Useful methods on the Application we are instrumenting
   class Application
-    # If the request is not valid for this service, we do not what to trace it.
-    def self.routable_request?(path_info, http_method)
+    # Does our framework know if this request will be routed to a controller?
```
ssteeg-mdsol: Maybe just me, but seeing a question sticks out to me like a TODO. How about:
```
Determines if our framework knows whether the request will be routed to a controller
```

lib/zipkin-tracer/rack/zipkin_env.rb
jcarres-mdsol: you are right, routable_request? is much more expensive so goes last

jcarres-mdsol: I see no need to update readme, I think it reflects reality already

PR #144, opened 2019-03-29
Record tags inconditionally at server receive

jcarres-mdsol:
This removes the possibility of constraining the tags recorded at server receive. It was always half-baked and pretty useless.

We originally created the feature of optionally setting tags in server because:
- UI bugs
- Being cheap with storage

I do not think any of these hold anymore.

Also, by our own experience using zipkin more often than not we want more info not less. 

This PR also adds `status_code` and `method` alongside the `path` info to server spans so that info is available when for some reason the client spans do not have it.

@adriancole @ykitamura-mdsol @jfeltesse-mdsol

### DIFF
--- a/lib/zipkin-tracer/config.rb
+++ b/lib/zipkin-tracer/config.rb
@@ -39,8 +39,6 @@ module ZipkinTracer
       if @sampled_as_boolean
         @logger && @logger.warn("Using a boolean in the Sampled header is deprecated. Consider setting sampled_as_boolean to false")
       end
-      # Record the given tags on server receive, even if the zipkin headers were present in the incoming request?
-      @record_on_server_receive = parse_tags(config[:record_on_server_receive])
 
       # When set to true, high 8-bytes will be prepended to trace_id.
       # The upper 4-bytes are epoch seconds and the lower 4-bytes are random.
@@ -73,14 +71,6 @@ module ZipkinTracer
       sampled_as_boolean: true,
       trace_id_128bit: false
     }
-
-    def parse_tags(tag_names)
-      return {} unless present?(tag_names)
-      names = tag_names.split(",").map(&:strip)
-      (ZipkinTracer::RackHandler::DEFAULT_SERVER_RECV_TAGS.keys & names).each_with_object({}) do |name, tags|
-        tags[name] = ZipkinTracer::RackHandler::DEFAULT_SERVER_RECV_TAGS[name]
-      end
-    end
 
     def present?(str)
       return false if str.nil?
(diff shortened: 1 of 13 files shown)


lib/zipkin-tracer/config.rb
```diff
@@ -39,8 +39,6 @@ def initialize(app, config_hash)
       if @sampled_as_boolean
         @logger && @logger.warn("Using a boolean in the Sampled header is deprecated. Consider setting sampled_as_boolean to false")
       end
-      # Record the given tags on server receive, even if the zipkin headers were present in the incoming request?
-      @record_on_server_receive = parse_tags(config[:record_on_server_receive])
```
ykitamura-mdsol: you can remove [`:record_on_server_receive` on line 11](https://github.com/openzipkin/zipkin-ruby/pull/144/files#diff-783b62518b0223696f3cc8e6c120d4e4R11)

lib/zipkin-tracer/rack/zipkin-tracer.rb
```diff
@@ -46,23 +47,17 @@ def annotate_plugin(span, env, status, response_headers, response_body)
...
-        @config.record_on_server_receive
-      elsif !zipkin_env.called_with_zipkin_headers?
-        # if the request comes from a non zipkin-enabled source record the default tags
-        DEFAULT_SERVER_RECV_TAGS
-      end
-      return if tags.nil?
-      tags.each { |annotation_key, env_key| span.record_tag(annotation_key, zipkin_env.env[env_key]) }
+    def trace_server_information(span, zipkin_env, status)
+      span.kind = Trace::Span::Kind::SERVER
+      span.record_tag(Trace::Span::Tag::STATUS, status)
```
adriancole: you can consider doubling this as error so the span appears red

ex from java
```java

    // 1xx, 2xx, and 3xx codes are not all valid, but the math is good enough vs drift and opinion
    // about individual codes in the range.
    if (httpStatusInt < 100 || httpStatusInt > 399) {
      customizer.tag("error", String.valueOf(httpStatusInt));
    }
```

ykitamura-mdsol: Here is how we are doing it in client spans:
https://github.com/openzipkin/zipkin-ruby/blob/cc4e358dbe8c6dab58d1aa62331d89f0ea42557e/lib/zipkin-tracer/faraday/zipkin-tracer.rb#L75

Maybe we can have a method like this in the [Span class](https://github.com/openzipkin/zipkin-ruby/blob/master/lib/zipkin-tracer/trace.rb#L166):
```ruby
    STATUS_ERROR_REGEXP = /\A(4.*|5.*)\z/.freeze

    def record_status(status)
      record_tag(Tag::STATUS, status)
      record_tag(Tag::ERROR, status) if STATUS_ERROR_REGEXP.match(status)
    end

jcarres-mdsol: Done

adriancole:
shibbit

PR #145, opened 2019-03-30
IP format is really a property of the tracer not of each endpoint. Move it over

jcarres-mdsol:
Also setup the default endpoint in config not in the tracer factory which should not manufacture other thing than tracers.

No functional changes

@ykitamura-mdsol @jfeltesse-mdsol @adriancole

### DIFF
--- a/lib/zipkin-tracer/zipkin_json_tracer.rb
+++ b/lib/zipkin-tracer/zipkin_json_tracer.rb
@@ -8,7 +8,8 @@ class AsyncJsonApiClient
   SPANS_PATH = '/api/v2/spans'
 
   def perform(json_api_host, spans)
-    spans_with_ips = ::ZipkinTracer::HostnameResolver.new.spans_with_ips(spans, ZipkinJsonTracer::IP_FORMAT).map(&:to_h)
+    spans_with_ips =
+      ::ZipkinTracer::HostnameResolver.new.spans_with_ips(spans, ZipkinJsonTracer::IP_FORMAT).map(&:to_h)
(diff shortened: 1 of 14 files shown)


lib/zipkin-tracer/zipkin_json_tracer.rb
```diff
@@ -8,7 +8,7 @@ class AsyncJsonApiClient
   SPANS_PATH = '/api/v2/spans'
 
   def perform(json_api_host, spans)
-    spans_with_ips = ::ZipkinTracer::HostnameResolver.new.spans_with_ips(spans).map(&:to_h)
+    spans_with_ips = ::ZipkinTracer::HostnameResolver.new.spans_with_ips(spans, ZipkinJsonTracer::IP_FORMAT).map(&:to_h)
```
ykitamura-mdsol: long line 🐶 

How about resolving spans in the `flush!` method like other tracers:
```ruby
  def flush!
    resolved_spans = ::ZipkinTracer::HostnameResolver.new.spans_with_ips(spans.dup, IP_FORMAT)
    AsyncJsonApiClient.perform_async(@json_api_host, resolved_spans)
  end
```

BTW, do you think we still need to do `spans.dup`?

jcarres-mdsol: `spans` are hold in a thread and wiped out if the thread is reused for a new request. I think it would be dangerous not to dup it.

Doing this operation in `perform` for performance reasons. You can dig up old PR for rationale, at that time performance skyrocketed.

PR #146, opened 2019-04-03
Fallback to empty string as Rails 5.1 recognize_path can't handle nil paths

ssteeg-mdsol:
#143 introduced an issue with Rails 5.1 (specifically actionpack), where passing an empty path to `Rails.application.routes.recognize_path` results in a `NoMethodError: undefined method encoding' for nil:NilClass` exception being thrown.

FYI: It's since been corrected in the rails code base - rails/rails@03925dc

I'm a bit unsure why this issue did not occur prior to openzipkin#143 though. Seems like the same problem should occur, but doesn't... 🤷‍♂️

@adriancole @JordiPolo @ykitamura-mdsol @jfeltesse-mdsol

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 0.35.1
+* Fix bug with Rails 5.1 when calling recognize_path with a nil path
+
 # 0.35.0
 * removes record_on_server_receive option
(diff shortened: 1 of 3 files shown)


jcarres-mdsol:
Nice and concise.

Issue #136, opened 2019-01-22
SQS integration


plevym:
Hi!

I have a rails app that consumes a queue using shoryuken (https://github.com/phstc/shoryuken), instead of receiving http requests. I've been trying to find a workaround so I can use zipkin to trace these, with no success. How can this be implemented?

devinsba:
The only zipkin implementation that I know of (in java) is: https://github.com/SmartThingsOSS/smartthings-brave/tree/master/brave-sqs

If you have the bandwidth maybe building something similar would be appreciated by the community!

PR #147, opened 2019-04-11
Add ZipkinSqsTracer to send spans via Amazon SQS

ykitamura-mdsol:
Added the [sender-sqs](https://github.com/openzipkin/zipkin-aws/tree/master/sender-sqs) like feature  to be able to send spans via Amazon SQS.

This is NOT solving the #136 issue.

### DIFF
--- a/lib/zipkin-tracer/zipkin_sqs_sender.rb
+++ b/lib/zipkin-tracer/zipkin_sqs_sender.rb
@@ -1,10 +1,10 @@
 require "aws-sdk-sqs"
 require "json"
-require "zipkin-tracer/zipkin_tracer_base"
+require "zipkin-tracer/zipkin_sender_base"
 require "zipkin-tracer/hostname_resolver"
 
 module Trace
-  class ZipkinSqsTracer < ZipkinTracerBase
+  class ZipkinSqsSender < ZipkinSenderBase
     IP_FORMAT = :string
 
     def initialize(options)
--- a/lib/zipkin-tracer/zipkin_sqs_sender.rb
+++ b/lib/zipkin-tracer/zipkin_sqs_sender.rb
@@ -1,29 +1,39 @@
 require "aws-sdk-sqs"
 require "json"
+require 'sucker_punch'
 require "zipkin-tracer/zipkin_sender_base"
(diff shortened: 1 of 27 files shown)


jfeltesse-mdsol:
This doesn't solve #136 because this PR allows you to _send_ spans when #136 is about _receiving_ spans, am I right?

adriancole:
ack. I would prefer using the term sender or reporter as tracer is usually discussing instrumentation

ykitamura-mdsol:
IIUC, #136 is to trace calls to SQS.

ykitamura-mdsol:
Agree. It is misleading to use the term `tracer` in these:
- tracer_factory.rb
- zipkin_json_tracer.rb
- zipkin_kafka_tracer.rb
- zipkin_sqs_tracer
- zipkin_logger_tracer
- zipkin_null_tracer

will try renaming them to `sender`

adriancole:
probably zipkin_json_tracer should be zipkin_http_sender

(or reporter doesn't matter)

ykitamura-mdsol:
renamed in https://github.com/openzipkin/zipkin-ruby/pull/147/commits/6dd9c4327dc4cf548dd9fa2cc96912f957e63610

lib/zipkin-tracer/zipkin_sqs_sender.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    def initialize(options)
+      sqs_options = options[:region] ? { region: options[:region] } : {}
+      @sqs = Aws::SQS::Client.new(**sqs_options)
+      @queue_name = options[:queue_name]
+      @logger = options[:logger]
+      super(options)
+    end
+
+    def flush!
```
jcarres-mdsol: This will be executed in the thread dealing with the request, maybe better to use the same Async mechanism than the http sender instead

ykitamura-mdsol: updated in https://github.com/openzipkin/zipkin-ruby/pull/147/commits/98c599b1d31c05f478136af504e8ae53dc7a9018

jcarres-mdsol:
Cool, nice new sender and good renaming, thanks!

PR #154, opened 2019-05-20
Annotate http.url instead of http.path

jcarres-mdsol:
I do not care about the host but I want the query parameters and could not find another standard annotation better than `url`. 
In many cases query parameters are key to understand why a particular request is slow.
For instance if you control pagination via query parameters you may be requesting a huge amount of data. Or if you do some complex filtering on your query params you may be requesting a very complex set that requires lots of processing.

@ykitamura-mdsol  @adriancole

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 0.37.0
+* Annotate http.url instead of http.path.
+
 # 0.36.2
 * Cleanup the gemspec. No code changes.
 
--- a/lib/zipkin-tracer/application.rb
+++ b/lib/zipkin-tracer/application.rb
@@ -4,8 +4,8 @@ module ZipkinTracer
     # Determines if our framework knows whether the request will be routed to a controller
     def self.routable_request?(env)
       return true unless defined?(Rails) # If not running on a Rails app, we can't verify if it is invalid
-      path_info = env[ZipkinTracer::RackHandler::PATH_INFO] || ""
-      http_method = env[ZipkinTracer::RackHandler::REQUEST_METHOD]
+      path_info = env[Rack::PATH_INFO] || ""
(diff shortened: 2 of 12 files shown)


ykitamura-mdsol:
One of our services is referring `path` info so this might be a breaking change for others too. 
How about keeping `path` as is and adding `query` with the value of `QUERY_STRING` in the Rack Env?

jcarres-mdsol:
We are all over the place but I think other traces tend to do `http.url` also so our other service may have prouble already

jcarres-mdsol:
@ykitamura-mdsol I'm going to do both url and path with a deprecation on path. Eventually to be removed

jcchavezs:
Honestly I would discourage you to do this change:
1. Query parameters are usually not annotated for privacy issues. People
could include private keys or access tokens into that and that is not
desired.
2. Annotating the full URL just because of the query parameters sound to me
something extreme. If there is a need for do such thing I would instead add
an option for passing a whitelist of query parameters you might want to
record under exceptional conditions and as a conscious decision to record
them.

Does any of the things I said make sense?



tir. 21. mai 2019, 00:11 skrev Jordi Polo Carres <notifications@github.com>:

jcarres-mdsol:
I could always add URL behind a configuration option and default it to not annotate it. In fact, thinking about this, I prefer that approach.

To me is crazy to have private info in query params but I guess the case does exist.

jcchavezs:
Yeah, old APIs use apitoken being passed in the query parameters.

In the other hand I like the idea is this being a optional feature at
client middleware level. What I am not sure is if the http.path should be
dropped when this option is enabled as `path` became more a `defacto`
annotation among libraries lately.

ons. 22. mai 2019, 02:35 skrev Jordi Polo Carres <notifications@github.com>:

PR #155, opened 2019-05-23
accept :async and :logger options, add a TraceWrapper utility

jfeltesse-mdsol:
See the README about the added/modified options and that utility class.

At the same time I add the standard `bin/` utilities that one would get from bundler upon generating a new gem skeleton.
`bin/console` makes quick prototyping easier.

@jcarres-mdsol @ykitamura-mdsol @ssteeg-mdsol @adriancole

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ Sidekiq.configure_server do |config|
 end
 ```
 
-By default workers aren't traced. you can specify the workers that you want to trace with traceable_workers config option. If you want all your workers to be traced pass [:all] to traceable_workers option (traceable_workers: [:all]).
+By default workers aren't traced. You can specify the workers that you want to trace with traceable_workers config option. If you want all your workers to be traced pass [:all] to traceable_workers option (traceable_workers: [:all]).
 
 ### Local tracing
 
@@ -224,7 +224,7 @@ lambda { |env| KNOWN_DEVICES.include?(env['HTTP_X_DEVICE_ID']) }
 This class provides a `.wrap_in_custom_span` method which expects a configuration hash, a span name and a block.
 You may also pass a span kind and an Application object using respectively `span_kind:` and `app:` keyword arguments.
 
-The block You pass will be executed in the context of a custom span.
+The block you pass will be executed in the context of a custom span.
 This is useful when your application doesn't use the rack handler but still needs to generate complete traces, for instance background jobs or lambdas calling remote services.
 
 The following code will create a trace starting with a span of the (default) `SERVER` kind named "custom span" and then a span of the `CLIENT` kind will be added by the Faraday middleware. Afterwards the configured sender will call `flush!`.
--- a/README.md
+++ b/README.md
@@ -8,27 +8,34 @@ Rack and Faraday integration middlewares for Zipkin tracing.
 
 ### Sending traces on incoming requests
 
-Options can be provided via Rails.config for a Rails 3+ app, or can be passed as a hash argument to the Rack plugin.
+Options can be provided as a hash via `Rails.config.zipkin_tracer` for Rails apps or directly to the Rack middleware:
 
 ```ruby
 require 'zipkin-tracer'
-use ZipkinTracer::RackHandler, config # config is optional
+use ZipkinTracer::RackHandler, config
 ```
 
-where `Rails.config.zipkin_tracer` or `config` is a hash that can contain the following keys:
+### Configuration options
 
-* `:service_name` **REQUIRED** - the name of the service being traced. There are two ways to configure this value. Either write the service name in the config file or set the "DOMAIN" environment variable (e.g. 'test-service.example.com' or 'test-service'). The environment variable takes precedence over the config file value.
-* `:sample_rate` (default: 0.1) - the ratio of requests to sample, from 0 to 1
-* `:json_api_host` - hostname with protocol of a zipkin api instance (e.g. `https://zipkin.example.com`) to use the HTTP sender
-* `:zookeeper` - the address of the zookeeper server to use by the Kafka sender
-* `:sqs_queue_name` - the name of the Amazon SQS queue to use the SQS sender
-* `:sqs_region` - the AWS region for the Amazon SQS queue
-* `:log_tracing` - Set to true to log all traces. Only used if traces are not sent to the API or Kafka.
-* `:annotate_plugin` - plugin function which receives the Rack env, the response status, headers, and body to record annotations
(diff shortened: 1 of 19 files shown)


lib/zipkin-tracer/trace_wrapper.rb
```diff
@@ -0,0 +1,18 @@
+module ZipkinTracer
+  class TraceWrapper
+    def self.wrap_in_custom_span(config, span_name, span_kind: Trace::Span::Kind::SERVER, app: nil)
```
ykitamura-mdsol: I think the span kind will be always `SERVER`, maybe it doesn't need to be passed as an argument... 🤔

lib/zipkin-tracer/trace_wrapper.rb
```diff
@@ -0,0 +1,18 @@
...
+      raise ArgumentError, "you must provide a block" unless block_given?
+
+      zipkin_config = ZipkinTracer::Config.new(app, config).freeze
+      tracer = ZipkinTracer::TracerFactory.new.tracer(zipkin_config)
+      trace_id = ZipkinTracer::TraceGenerator.new.next_trace_id
+
+      ZipkinTracer::TraceContainer.with_trace_id(trace_id) do
+        tracer.with_new_span(trace_id, span_name) do |span|
```
ykitamura-mdsol: We need to check `sampled?` here.
Also, some might want to add additional information like in [local-tracing](https://github.com/openzipkin/zipkin-ruby#local-tracing).
How about passing `span` to the `yield`?

```ruby
if trace_id.sampled?
  tracer.with_new_span(trace_id, span_name) do |span|
    span.kind = span_kind
    yield(span)
  end
else
  yield(ZipkinTracer::NullSpan.new)
end
```

lib/zipkin-tracer/trace_wrapper.rb
jfeltesse-mdsol: That's my thinking too but allowing for it to be passed gives the extra flexibility at no cost.
We've already seen people doing wonky things with the library so might be someone will come up with a specific use case where they need to customize the kind.

README.md
```diff
@@ -8,27 +8,34 @@ Rack and Faraday integration middlewares for Zipkin tracing.
...
-* `:annotate_plugin` - plugin function which receives the Rack env, the response status, headers, and body to record annotations
-* `:filter_plugin` - plugin function which receives the Rack env and will skip tracing if it returns false
-* `:whitelist_plugin` - plugin function which receives the Rack env and will force sampling if it returns true
-* `:sampled_as_boolean` - When set to true (default but deprecrated), it uses true/false for the `X-B3-Sampled` header. When set to false uses 1/0 which is preferred.
-* `:trace_id_128bit` - When set to true, high 8-bytes will be prepended to trace_id. The upper 4-bytes are epoch seconds and the lower 4-bytes are random. This makes it convertible to Amazon X-Ray trace ID format v1. (See http://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-request-tracing.html)
+#### Common
+- `:service_name` **REQUIRED** - the name of the service being traced. There are two ways to configure this value. Either write the service name in the config file or set the "DOMAIN" environment variable (e.g. 'test-service.example.com' or 'test-service'). The environment variable takes precedence over the config file value.
```
ssteeg-mdsol: [nit] Why the change from `*` → `-`? The "Plugins" section is also still using `*`

README.md
```diff
@@ -210,6 +217,29 @@ For example:
...
 ```
 
+## Utility classes
+
+### TraceWrapper
+
+This class provides a `.wrap_in_custom_span` method which expects a configuration hash, a span name and a block.
+You may also pass a span kind and an Application object using respectively `span_kind:` and `app:` keyword arguments.
+
+The block You pass will be executed in the context of a custom span.
```
ssteeg-mdsol: The block ~Y~you pass

README.md
```diff
@@ -210,6 +217,29 @@ For example:
...
+This is useful when your application doesn't use the rack handler but still needs to generate complete traces, for instance background jobs or lambda calling remote services.
```
ssteeg-mdsol: lambda → lambdas

README.md
jfeltesse-mdsol: 🙈 
just a habit. 

fixed in 698db93e6586dc8ebd33f2ff983a73e69e665785

lib/zipkin-tracer/trace_wrapper.rb
jfeltesse-mdsol: Great suggestions! Implemented in 8598529a56039cb08be9147b5a9aca3f0d230fcd

jcarres-mdsol:
I guess similar code would be needed for other tracers using async, maybe can be isolated in a module or class. Why not make the change to all of them so we do not have the caviat of the that option only working on one?

Also, the sidekiq integration could be written now on top of the wrapper class?

jfeltesse-mdsol:
> I guess similar code would be needed for other tracers using async, maybe can be isolated in a module or class. Why not make the change to all of them so we do not have the caviat of the that option only working on one?

time limit didn't allow me to go further.

> Also, the sidekiq integration could be written now on top of the wrapper class?

the topic of another PR, surely.

jcarres-mdsol:
At least I think "async option to be applied to all traces which do async"  (maybe just one more tracer)  should be implemented in this PR, just to not leave it hanging. I see the point about sidekick

jfeltesse-mdsol:
@jcarres-mdsol alright, done in 013332f. Enjoy!

lib/zipkin-tracer/zipkin_http_sender.rb
```diff
@@ -32,13 +32,18 @@ class ZipkinHttpSender < ZipkinSenderBase
...
+      SuckerPunch.logger = options[:logger]
       super(options)
     end
 
     def flush!
-      AsyncHttpApiClient.perform_async(@json_api_host, spans.dup)
+      if @async
+        HttpApiClient.perform_async(@json_api_host, spans.dup)
+      else
+        HttpApiClient.new.perform(@json_api_host, spans.dup)
```
jcarres-mdsol: I think we do not need to `.dup` here. Same comment for the other sender.
The reason we dup in async is to avoid the thread variable to be overwritten as the thread gets a new request while async is still to sent the spans of the previous request. In the sync case, that should not be an issue.

jfeltesse-mdsol: done in f60e276

PR #156, opened 2019-07-10
Add rabbit mq sender

sonny8988:
I would like to use this gem with rabbitmq as the transport.
This PR adds this support for the gem.

### DIFF
--- a/lib/zipkin-tracer/zipkin_rabbit_mq_sender.rb
+++ b/lib/zipkin-tracer/zipkin_rabbit_mq_sender.rb
@@ -1,18 +1,22 @@
 require 'json'
+require 'sucker_punch'
 require 'zipkin-tracer/zipkin_sender_base'
 require 'zipkin-tracer/hostname_resolver'
 
 module Trace
   class RabbitMqPublisher
-    def initialize(connection, exchange, routing_key)
-      @connection = connection
-      channel = @connection.create_channel
-      @exchange = channel.exchange(exchange)
-      @routing_key = routing_key
-    end
+    include SuckerPunch::Job
 
-    def publish(message)
-      @exchange.publish(message, routing_key: @routing_key)
+    def perform(exchange, routing_key, spans)
+      spans_with_ips = ::ZipkinTracer::HostnameResolver.new
+        .spans_with_ips(spans, ZipkinRabbitMqSender::IP_FORMAT)
+        .map(&:to_h)
+
+      message = JSON.generate(spans_with_ips)
+
+      exchange.publish(message, routing_key: routing_key)
+    rescue => e
+      SuckerPunch.logger.error(e)
     end
   end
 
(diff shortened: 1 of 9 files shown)


sonny8988:
Looks like there is an issue with rvm jruby

spec/lib/zipkin_rabbitmq_sender_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+    let(:name) { 'test' }
+    let(:span) { tracer.start_span(trace_id, name) }
+    let(:spans) do
+      ::ZipkinTracer::HostnameResolver.new
+        .spans_with_ips([span], described_class::IP_FORMAT)
+        .map(&:to_h)
+    end
+
+    before do
+      Timecop.freeze
```
adriancole: it would be nice for at least one test to show message literal instead of checking variables. this can make clear for example it is a list of json v2 format as opposed to v1

sonny8988: Sure, I updated the tests.

lib/zipkin-tracer/zipkin_rabbit_mq_sender.rb
```diff
@@ -0,0 +1,42 @@
...
+
+module Trace
+  class RabbitMqPublisher
+    def initialize(connection)
+      @connection = connection
+      @channel = @connection.create_channel
+    end
+
+    def publish(exchange, routing_key, message)
+      exchange = @channel.exchange(exchange)
```
jcarres-mdsol: Does this operation needs to be done on each `publish`? Both @channel and @exchange does not seem to change.

lib/zipkin-tracer/zipkin_rabbit_mq_sender.rb
```diff
@@ -0,0 +1,42 @@
...
+      exchange.publish(message, routing_key: routing_key)
```
jcarres-mdsol: If this is a slow operation the thread on your application working on this will stop till it is done.
Consider , if it makes sense, to make it async (you can copy from how json sender is doing it)

jcarres-mdsol:
Couple of comments but it is great overall!!

lib/zipkin-tracer/zipkin_rabbit_mq_sender.rb
sonny8988: I guess it doesn't really need to be done on each publish, I moved it out.

lib/zipkin-tracer/zipkin_rabbit_mq_sender.rb
sonny8988: I rewrote the publisher using suckerpunch to make it async similar to the other senders.

jcarres-mdsol:
Great. Thanks!

PR #158, opened 2019-07-17
denote the RabbitMQ sender supports the :async parameter

jfeltesse-mdsol:
all in the title.

@jcarres-mdsol 	@ssteeg-mdsol 	@adriancole


jcarres-mdsol:
Thanks!

PR #159, opened 2019-07-23
Allow clients to pass in explicit start and end timestamps to spans

jvans1:
The [open tracing spec](https://github.com/opentracing/specification/blob/master/specification.md) allows for the clients to pass an optional start and end timestamp to spans. This change enables that option for clients while providing the default of the current wall time as defined in the spec.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: ruby
+dist: trusty
 jdk:
   - oraclejdk8
 rvm:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 0.39.0
+* Allow clients to provide an explicit timestamp when starting and stopping spans
+
 # 0.38.0
 * Add RabbitMQ sender
 
--- a/lib/zipkin-tracer/trace.rb
+++ b/lib/zipkin-tracer/trace.rb
@@ -179,7 +179,7 @@ module Trace
 
     attr_accessor :name, :kind, :local_endpoint, :remote_endpoint, :annotations, :tags, :debug
 
-    def initialize(name, span_id)
+    def initialize(name, span_id, timestamp = Time.now)
       @name = name
       @span_id = span_id
       @kind = nil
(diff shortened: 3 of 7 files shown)


adriancole:
opentracing is dead, but anyway this isn't an opentracing project. please don't cite OT as rationale for anything.

What are you aiming to do with this? Sometimes people are trying to record spans after the fact, and there are better ways to do that.

jvans1:
Sorry I didn't realize this wasn't an OT implementation, a lot of the vernacular is the same but I guess they differ in some ways. What is the best way to record a span after the fact? I've dug into most of the code and had a hard time finding a clean way to do that.

adriancole:
I meant to use the sender directly, like this for the out-of-band spans

https://github.com/openzipkin/zipkin-ruby/blob/master/lib/zipkin-tracer/zipkin_http_sender.rb

I'm not sure how much lifecycle is implicit in the sender api. In other projects they just take object literals, but this one seems to invoke lifecycle unless I'm mistaken @ykitamura-mdsol 
https://github.com/openzipkin/zipkin-ruby/blob/9cc7ba6b42955e7f7fff9528c40d90e8071508a1/lib/zipkin-tracer/zipkin_sender_base.rb

anyway it would seem not a terribly hard refactor to allow senders to accept complete span objects, ideally plain ones like structs

jcarres-mdsol:
@jvans1 still interested in this? I'd merge, add a line to the changelog and update the version

jvans1:
@jcarres-mdsol yes! I've added a line to the changelog and bumped the version. I noticed the travis build was failing and it looks related to some [openjdk issue](https://travis-ci.community/t/error-installing-oraclejdk8-expected-feature-release-number-in-range-of-9-to-14-but-got-8/3766/6) on travis. I added their recommended fix of specifying `dist: trusty` in the travis.yml and it seems to have fixed it

jcarres-mdsol:
Looks great. Thanks!

jcarres-mdsol:
Thanks for fixing the failing build !

PR #160, opened 2019-08-20
Duplicate rack environment strings before calling recognize

jvans1:
Certain gems like [routing-filter](https://github.com/svenfuchs/routing-filter) modify the rack environment in place when recognizing a route. Applications that use this type of behavior depend on recognize only being called once. Third parties that use this call break these types of applications if they call recognize without duping the rack variables.

I couldn't really think of a better way to test this without the custom matcher but I'm all ears if someone has a better way to do it!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 0.39.1
+* Duplicate rack environment strings so that applications that depend on this call only happening once don't break
+
 # 0.39.0
 * Allow clients to provide an explicit timestamp when starting and stopping spans
 
--- a/lib/zipkin-tracer/application.rb
+++ b/lib/zipkin-tracer/application.rb
@@ -12,13 +12,16 @@ module ZipkinTracer
       false
     end
 
+    def self.stub_env(env)
+      {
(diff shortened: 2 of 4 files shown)


jcarres-mdsol:
I'd ask review from @jfeltesse-mdsol and @ykitamura-mdsol , they may have some idea about the test. 
Not a nice situation :/

ykitamura-mdsol:
I can only think of separating the logic to a method and test against it...

e.g.
```ruby
def self.stub_env(env)
  {
    "PATH_INFO" => env[ZipkinTracer::RackHandler::PATH_INFO].dup,
    "REQUEST_METHOD" => env[ZipkinTracer::RackHandler::REQUEST_METHOD].dup
  }
end
```

```ruby
let(:stub_env) { Application.stub_env(env) }

%w(PATH_INFO REQUEST_METHOD).each do |field|
  expect(stub_env[field]).to eq(env[field])
  expect(stub_env[field]).not_to equal(env[field])
end
```

jvans1:
@ykitamura-mdsol good suggestion, thanks. I've updated the PR with your change

jfeltesse-mdsol:
LGTM

PR #162, opened 2019-09-07
protect against double inserting middlewares

jcarres-mdsol:
It seems that mistake do happen and does not look nice. 
Thoughts on the approach? I could not find a way to find out at initialization time.

@ykitamura-mdsol  @jfeltesse-mdsol

### DIFF
--- a/lib/zipkin-tracer/faraday/zipkin-tracer.rb
+++ b/lib/zipkin-tracer/faraday/zipkin-tracer.rb
@@ -10,6 +10,8 @@ module ZipkinTracer
     end
 
     def call(env)
+      # Prevent calling this middleware twice
+      return @app.call(env) if !env[:request_headers][b3_headers().values.first].nil?
       trace_id = TraceGenerator.new.next_trace_id
       TraceContainer.with_trace_id(trace_id) do
         b3_headers.each do |method, header|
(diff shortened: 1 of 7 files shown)


ykitamura-mdsol:
Why not just fix the mistake?

jfeltesse-mdsol:
is this to prevent end users from shooting themselves in the foot by adding the middleware twice in their chain?

jcarres-mdsol:
Right, is to prevent users from adding it twice.
I've seen this happen in one instance, could fix it there also and there is an argument for doing that instead. But on the other hand do not want to play a whac-a-mole game. What do you think?

spec/lib/excon/zipkin-tracer_spec.rb
```diff
@@ -15,6 +15,7 @@ def process(body, url, headers = {})
                           )
     connection.request
 
+    return headers if !headers.empty?
```
jfeltesse-mdsol: `unless headers.empty?`

lib/zipkin-tracer/faraday/zipkin-tracer.rb
```diff
@@ -10,6 +10,8 @@ def initialize(app, service_name = nil)
     end
 
     def call(env)
+      # Prevent calling this middleware twice
+      return @app.call(env) if !env[:request_headers][b3_headers().values.first].nil?
```
jfeltesse-mdsol: - How about printing some warning in the "called twice" case? Silently returning means the user will never know their config is dodgy.
- Code-wise how about making the headers names constants and simply pick one directly, e.g. 
  `return @app.call(env) if env[:request_headers].key?(X-B3-TRACEID)`. 
  The current implementation here will create a new array from the hash values on every single call, even when the config is sane.

PR #165, opened 2019-10-08
Add support for reading B3 single header

ykitamura-mdsol:
This PR is just to support the "reading" part of B3 single header (#126).
The writing part is not included.

@adriancole sorry for taking a long time but I like the single header!

@jcarres-mdsol @jfeltesse-mdsol

### DIFF
--- a/lib/zipkin-tracer/rack/zipkin_env.rb
+++ b/lib/zipkin-tracer/rack/zipkin_env.rb
@@ -10,10 +10,14 @@ module ZipkinTracer
     end
 
     def trace_id(default_flags = Trace::Flags::EMPTY)
-      trace_id, span_id, parent_span_id, shared = retrieve_or_generate_ids
-      sampled = sampled_header_value(@env['HTTP_X_B3_SAMPLED'])
-      flags = (@env['HTTP_X_B3_FLAGS'] || default_flags).to_i
+      trace_id, span_id, parent_span_id, sampled, flags, shared = retrieve_or_generate_ids
+      sampled = sampled_header_value(sampled)
+      flags = (flags || default_flags).to_i
       Trace::TraceId.new(trace_id, parent_span_id, span_id, sampled, flags, shared)
+    end
+
+    def called_with_zipkin_b3_single_header?
+      @called_with_zipkin_b3_single_header ||= @env.key?(B3_SINGLE_HEADER)
     end
 
     def called_with_zipkin_headers?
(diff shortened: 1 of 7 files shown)


lib/zipkin-tracer/zipkin_b3_single_header_format.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    def to_a
+      [trace_id, span_id, parent_span_id, sampled, flags]
+    end
+
+    def self.parse_from_header(b3_single_header)
+      return new(parse_sampled_flags(b3_single_header)) if b3_single_header.size == 1
+
+      trace_id, span_id, flag, parent_span_id = b3_single_header.split('-')
+      new(trace_id: trace_id, span_id: span_id, parent_span_id: parent_span_id, **parse_sampled_flags(flag))
```
jfeltesse-mdsol: I think it's a bit wasteful to have this method create a new array (by calling split), then a new object and then in the caller call `to_a` which is going to create yet another array.

How about having a method getting the header as an input and return an array directly, without instantiating?

lib/zipkin-tracer/zipkin_b3_single_header_format.rb
```diff
@@ -0,0 +1,38 @@
+module ZipkinTracer
```
jfeltesse-mdsol: how about adding the magic comment about frozen strings?

lib/zipkin-tracer/rack/zipkin_env.rb
```diff
@@ -10,33 +10,44 @@ def initialize(env, config)
...
+
     def called_with_zipkin_headers?
       @called_with_zipkin_headers ||= B3_REQUIRED_HEADERS.all? { |key| @env.key?(key) }
     end
 
     private
 
-    B3_REQUIRED_HEADERS = %w(HTTP_X_B3_TRACEID HTTP_X_B3_SPANID).freeze
-    B3_OPT_HEADERS = %w(HTTP_X_B3_PARENTSPANID HTTP_X_B3_SAMPLED HTTP_X_B3_FLAGS).freeze
+    B3_SINGLE_HEADER = 'HTTP_B3'.freeze
```
adriancole: easier to quote, ain't it :)

lib/zipkin-tracer/zipkin_b3_single_header_format.rb
```diff
@@ -0,0 +1,38 @@
...
+  # This format corresponds to the propagation key "b3" (or "B3").
+  # b3: {x-b3-traceid}-{x-b3-spanid}-{if x-b3-flags 'd' else x-b3-sampled}-{x-b3-parentspanid}
+  # For details, see: https://github.com/apache/incubator-zipkin-b3-propagation
```
adriancole:
```suggestion
  # For details, see: https://github.com/openzipkin/b3-propagation
```

lib/zipkin-tracer/zipkin_b3_single_header_format.rb
adriancole: though I would say this is the least code approach I've ever seen :P

spec/lib/zipkin_b3_single_header_format_spec.rb
```diff
@@ -0,0 +1,113 @@
+require 'spec_helper'
+
+describe ZipkinTracer::B3SingleHeaderFormat do
+  let(:b3_single_header_format) { described_class.parse_from_header(b3_single_header) }
+
+  context 'child span' do
```
adriancole: add some malformed tests? Ex in brave we use

"not-a-tumor"
"b970dafd-0d95-40aa-95d8-1d8725aebe40"

lib/zipkin-tracer/zipkin_b3_single_header_format.rb
ykitamura-mdsol: Added in f8d6a89

lib/zipkin-tracer/zipkin_b3_single_header_format.rb
ykitamura-mdsol: Updated in 2fef91b

spec/lib/zipkin_b3_single_header_format_spec.rb
ykitamura-mdsol: @adriancole In ruby client, we are doing malformed check in another class: 

https://github.com/openzipkin/zipkin-ruby/blob/a1b1523a5bfcde3b6d2a8213a76e76b7a0e8a088/lib/zipkin-tracer/trace.rb#L134-L150

I think we don't need to have malformed tests here but what do you think?

adriancole:
sounds fine

PR #166, opened 2019-10-14
Add support for writing B3 single header

### LINKED ISSUE #115: Should omit ParentSpanId header for root spans
There are services which throw error when an empty `X-B3-ParentSpanId` header sent with the HTTP request.

From the [openzipkin specification](https://github.com/openzipkin/b3-propagation#parentspanid-1):

> The `X-B3-ParentSpanId` header must be present on a child span and absent on the root span.

## How to reproduce

```ruby
conn = Faraday.new() do |faraday|
  faraday.use ZipkinTracer::FaradayHandler, 'service_name'
  faraday.request :url_encoded
  faraday.adapter Faraday.default_adapter
end
conn.get('http://logs-01.loggly.com/inputs/TOKEN/tag/http/')
```

Result is `400 Bad request`.

If we don't send tracing headers, the response is the expected `403 Invalid API key`:
```ruby
conn = Faraday.new() do |faraday|
  faraday.request :url_encoded
  faraday.adapter Faraday.default_adapter
end
conn.get('http://logs-01.loggly.com/inputs/TOKEN/tag/http/')
```

## Workaround

The only workaround i found so far is to monkey patch FaradayHandler with `unless trace_id.send(method).to_s.empty?` as below:

```ruby
require 'zipkin-tracer/faraday/zipkin-tracer'
module ZipkinTracer
  class FaradayHandler < ::Faraday::Middleware
    def call(env)
      trace_id = TraceGenerator.new.next_trace_id
      TraceContainer.with_trace_id(trace_id) do
        b3_headers.each do |method, header|
          env[:request_headers][header] = trace_id.send(method).to_s unless trace_id.send(method).to_s.empty?
        end
        if Trace.tracer && trace_id.sampled?
          trace!(env, trace_id)
        else
          @app.call(env)
        end
      end
    end
  end
end
```

ykitamura-mdsol:
This PR is to support the "writing" part of B3 single header (#126)

Also, fixes #115

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 * Add support for writing B3 single header.
 * Omit ParentSpanId header for root spans.
 
+# 0.40.1
+* Fix to pass `async` option to the HTTP sender.
+
 # 0.40.0
 * Add support for reading B3 single header.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 0.41.0
+* Add support for writing B3 single header.
+* Omit ParentSpanId header for root spans.
+
 # 0.40.0
 * Add support for reading B3 single header.
 
--- a/lib/zipkin-tracer/config.rb
+++ b/lib/zipkin-tracer/config.rb
@@ -8,7 +8,7 @@ module ZipkinTracer
     attr_reader :service_name, :sample_rate, :sampled_as_boolean, :trace_id_128bit, :async, :logger,
       :json_api_host, :zookeeper, :kafka_producer, :kafka_topic, :sqs_queue_name, :sqs_region, :log_tracing,
       :annotate_plugin, :filter_plugin, :whitelist_plugin, :rabbit_mq_connection, :rabbit_mq_exchange,
-      :rabbit_mq_routing_key
+      :rabbit_mq_routing_key, :write_b3_single_format
 
     def initialize(app, config_hash)
       config = config_hash || Application.config(app)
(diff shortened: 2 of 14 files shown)


spec/lib/middleware_shared_examples.rb
```diff
@@ -91,18 +93,33 @@ def expect_tracing
...
+          request_headers = process('', url)
+        end
 
-      expect(request_headers['X-B3-TraceId']).to eq('0000000000000001')
-      expect(request_headers['X-B3-ParentSpanId']).to eq('0000000000000003')
-      expect(request_headers['X-B3-SpanId']).not_to eq('0000000000000003')
-      expect(request_headers['X-B3-SpanId']).to match(HEX_REGEX)
-      expect(request_headers['X-B3-Sampled']).to eq('true')
-      expect(request_headers['X-B3-Flags']).to eq('1')
+        expect(request_headers['B3']).to match(/\A0000000000000001-\h{16}-d-0000000000000003\z/)
```
adriancole: please downcase this by default.

lib/zipkin-tracer/config.rb
```diff
@@ -53,9 +53,13 @@ def initialize(app, config_hash)
       # This makes it convertible to Amazon X-Ray trace ID format v1.
       # (See http://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-request-tracing.html)
       @trace_id_128bit = config[:trace_id_128bit].nil? ? DEFAULTS[:trace_id_128bit] : config[:trace_id_128bit]
+      # When set to true, only writes a single b3 header for outbound propagation.
+      @write_b3_single_format =
+        config[:write_b3_single_format].nil? ? DEFAULTS[:write_b3_single_format] : config[:write_b3_single_format]
```
jfeltesse-mdsol: Is there a glimpse of a convention around naming configuration options across zipkin libraries?
My reasoning is the name could be better but if the other libraries implementing that header format have gone with this one then it's better to conform.

About the name itself, I think it doesn't convey the idea unless you've learned about the "classic" headers and that new "single B3" header beforehand.
I don't know if appropriate but something like "single_b3_header" would be easier to grok IMHO.

spec/lib/middleware_shared_examples.rb
ykitamura-mdsol: updated in 00f824b

jfeltesse-mdsol:
can you squash the commits in your fork so the history doesn't feature those extra commits related to syncing with upstream?

ykitamura-mdsol:
Rebased on the top of `0.40.1`

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+# 0.41.0
+* Add support for writing B3 single header.
+* Omit ParentSpanId header for root spans.
+
+# 0.40.1
+* Fix to pass `async` option to the HTTP sender.
```
jfeltesse-mdsol: sorry for not picking this up in #167 and thanks for adding it 🙇

PR #167, opened 2019-10-21
option :async is not passed to http sender

turnon:
v0.37.0 said it added an `async` option to the HTTP and SQS senders, but in fact only added to SQS. This PR fix it.

### DIFF
--- a/lib/zipkin-tracer/tracer_factory.rb
+++ b/lib/zipkin-tracer/tracer_factory.rb
@@ -6,7 +6,11 @@ module ZipkinTracer
       tracer = case adapter
         when :json
           require 'zipkin-tracer/zipkin_http_sender'
-          options = { json_api_host: config.json_api_host, logger: config.logger }
+          options = {
+            async: config.async,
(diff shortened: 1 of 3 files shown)


lib/zipkin-tracer/tracer_factory.rb
```diff
@@ -6,7 +6,11 @@ def tracer(config)
       tracer = case adapter
         when :json
           require 'zipkin-tracer/zipkin_http_sender'
-          options = { json_api_host: config.json_api_host, logger: config.logger }
+          options = {
+            json_api_host: config.json_api_host,
+            async: config.async,
```
jfeltesse-mdsol: can you puts async first so the order is alphabetical (like the SQS sender)?

turnon: ok

turnon:
> Don't remember so much the code but would it be easy to add specs to ensure that switch case statement passes the expected options to the correct initializer?

specs added

jcarres-mdsol:
Thanks!

PR #170, opened 2020-01-26
Add PRODUCER and CONSUMER span kinds

ykitamura-mdsol:
Added the `PRODUCER` and `CONSUMER` span kinds.

@adriancole @jcarres-mdsol @bvillanueva-mdsol

### DIFF
--- a/lib/zipkin-tracer/trace.rb
+++ b/lib/zipkin-tracer/trace.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 require 'zipkin-tracer/zipkin_sender_base'
 require 'zipkin-tracer/trace_container'
 # Most of this code is copied from Finagle
@@ -165,16 +167,16 @@ module Trace
   # A span may contain many annotations
   class Span
     module Tag
-      METHOD = "http.method".freeze
-      PATH = "http.path".freeze
-      STATUS = "http.status_code".freeze
-      LOCAL_COMPONENT = "lc".freeze # TODO: Remove LOCAL_COMPONENT and related methods when no longer needed
-      ERROR = "error".freeze
+      METHOD = "http.method"
+      PATH = "http.path"
+      STATUS = "http.status_code"
+      LOCAL_COMPONENT = "lc" # TODO: Remove LOCAL_COMPONENT and related methods when no longer needed
+      ERROR = "error"
     end
 
     module Kind
-      CLIENT = "CLIENT".freeze
-      SERVER = "SERVER".freeze
+      CLIENT = "CLIENT"
+      SERVER = "SERVER"
 
       # When present, "timestamp" is the moment a producer sent a message to a destination.
       # "duration" represents delay sending the message, such as batching, while
@@ -182,7 +184,7 @@ module Trace
       #
       # Unlike CLIENT, messaging spans never share a span ID. For example, the
       # CONSUMER of the same message has "parent_id" set to this span's id.
-      PRODUCER = "PRODUCER".freeze
+      PRODUCER = "PRODUCER"
 
       # When present, "timestamp" is the moment a consumer received a message from an origin.
       # "duration" represents delay consuming the message, such as from backlog,
@@ -190,7 +192,7 @@ module Trace
       #
       # Unlike SERVER, messaging spans never share a span ID. For example, the
       # PRODUCER of this message is the "parent_id" of this span.
-      CONSUMER = "CONSUMER".freeze
+      CONSUMER = "CONSUMER"
     end
 
     attr_accessor :name, :kind, :local_endpoint, :remote_endpoint, :annotations, :tags, :debug
--- a/lib/zipkin-tracer/trace.rb
+++ b/lib/zipkin-tracer/trace.rb
@@ -195,8 +195,6 @@ module Trace
       CONSUMER = "CONSUMER"
     end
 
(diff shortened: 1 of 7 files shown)


spec/lib/zipkin_sender_base_spec.rb
```diff
@@ -76,17 +76,16 @@
...
-    end
-    it 'flush if kind is CLIENT in this span and the span does not have parent span' do
-      span.kind = Trace::Span::Kind::CLIENT
-      expect(tracer).to receive(:flush!)
-      expect(tracer).to receive(:reset)
-      tracer.end_span(span)
+    [
+      Trace::Span::Kind::SERVER, Trace::Span::Kind::CLIENT,
+      Trace::Span::Kind::PRODUCER, Trace::Span::Kind::CONSUMER
+    ].each do |kind|
```
jfeltesse-mdsol: any reason not to go with rspec's [shared examples](https://relishapp.com/rspec/rspec-core/v/3-9/docs/example-groups/shared-examples)?

lib/zipkin-tracer/trace.rb
```diff
@@ -175,6 +175,22 @@ module Tag
     module Kind
       CLIENT = "CLIENT".freeze
       SERVER = "SERVER".freeze
+
+      # When present, "timestamp" is the moment a producer sent a message to a destination.
+      # "duration" represents delay sending the message, such as batching, while
+      # "remote_endpoint" indicates the destination, such as a broker.
+      #
+      # Unlike CLIENT, messaging spans never share a span ID. For example, the
+      # CONSUMER of the same message has "patient_id" set to this span's id.
```
adriancole: medical brains?

```suggestion
      # CONSUMER of the same message has "parent_id" set to this span's id.
```

lib/zipkin-tracer/zipkin_sender_base.rb
```diff
@@ -25,7 +25,7 @@ def end_span(span, timestamp = Time.now)
       # If in a thread not handling incoming http requests, it will not have Kind::SERVER, so the span
       # will never be flushed and will cause memory leak.
       # If no parent span, then current span needs to flush when it ends.
-      if !span.has_parent_span? || span.kind == Trace::Span::Kind::SERVER
+      if !span.has_parent_span? || span.kind == Trace::Span::Kind::SERVER || span.kind == Trace::Span::Kind::CONSUMER
```
adriancole: I don't fully understand the history of this, so not sure if this will be relevant in the case of consumer span. another check you should do is if any code would result in a consumer span sharing the same id as the producer.. it should never do that (and therefore never also set shared=true)

ykitamura-mdsol: Probably #49 explains about the history, we don't have buffering so I think we need to `flush` when span kind is `CONSUMER`.
@jcarres-mdsol Please correct me if I'm misunderstanding.

> another check you should do is if any code would result in a consumer span sharing the same id as the producer.. it should never do that (and therefore never also set shared=true)

I don't have a good idea of how we can perform this check in the ruby client but is it okay that we make sure to always generate a new span id for producer and consumer?

lib/zipkin-tracer/trace.rb
ykitamura-mdsol: 🙈

lib/zipkin-tracer/trace.rb
```diff
@@ -175,6 +175,22 @@ module Tag
...
+      PRODUCER = "PRODUCER".freeze
```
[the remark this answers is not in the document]
ykitamura-mdsol: updated in ae0fb7d

spec/lib/zipkin_sender_base_spec.rb
ykitamura-mdsol: updated in 8859e64

lib/zipkin-tracer/zipkin_sender_base.rb
jcarres-mdsol: I think the original code really is trying to flush (send to server and remove local stack) at the end of the request or if you are in some kind of backgground process then all the time, you are not so time constraint. 
In that sense PRODUCER and CONSUMER would probably happen on background jobs, shouldn't both trigger a flush?

adriancole: in instrumentation I've worked on, consumer is a poll loop and so flush isn't an issue more a span.finish command when you are done. It is probably easier to see the impact of these things with instrumentation that use them.

pps yeah always forking on consumer is good, just mean also we can add a unit test that never shared=true if unsure where that occurs in the codebase. If we are sure where that occurs and we know it won't be set shared=true, then no need for unit test

my 2p

jcarres-mdsol: Right, IIRC, to allow us to create local spans and what not without over-flushing, our span.finish does not flush, it keeps data in a thread local storage. So flush needs to happen at some point and the logic we are looking at is trying to be clever at when that is.

ykitamura-mdsol: "shared" can be configured manually so updated in cd54388 to send only in the SERVER and CLIENT spans

adriancole: actually only SERVER, never client or anything else.

PR #171, opened 2020-02-13
Fix to not flush producer span if a server span is in the stack

ykitamura-mdsol:
A producer span can be created within a server span and if we flush the producer span before the server span gets closed, the server span will lose its duration information.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 0.43.1
+* Fix to not flush `PRODUCER` span when a server span is in the stack.
+
 # 0.43.0
 * Add the `PRODUCER` and `CONSUMER` span kinds.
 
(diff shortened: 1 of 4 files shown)


adriancole:
I get the what and why, and agree this should be fixed.

.. just both are getting gnarly :) Would be nice to have a mechanism that has less maintenance involved at some point.

ykitamura-mdsol:
> .. just both are getting gnarly :) Would be nice to have a mechanism that has less maintenance involved at some point.

How about flushing when the root (oldest) span gets closed?
```ruby
return unless spans.first.equal?(span)
```

adriancole:
it might be too much, but something to consider.. in brave, we add another data to the internal model called localRootId.

https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/propagation/TraceContext.java#L120

This extra field would be set at the entrypoint, whether that is from a message consumer or a server. This allows a sort-of group-by or sub-tree, as you can hold back on whichever is the localRoot. This would be more involved, but could keep what I think is the intent here. to block until a certain subtree are done.

PR #173, opened 2020-04-02
Allow Faraday 1.x

jfeltesse-mdsol:
The bulk of the "work" here was in the integration TestApp helper. It hanged for some reason (race?), and in any case the code was convoluted. After refactoring, no issues.

@adriancole @jcarres-mdsol @ykitamura-mdsol


jfeltesse-mdsol:
Follow up question: should we push faraday 1.x or make it accept both 0.x or 1.x? I didn't go that route because we'd then need to test both versions using appraisal or wwtd.

jcarres-mdsol:
I would imagine there is an universe of things out there depending on Faraday < 1 , would it take a while for libraries and applications to move to faraday 1.x?

jfeltesse-mdsol:
@jcarres-mdsol added Appraisal to test both 0.x and 1.x. Should be good to go.

ykitamura-mdsol:
looks good to me 👍

.travis.yml
```diff
@@ -18,12 +25,12 @@ deploy:
   on:
     tags: true
     repo: openzipkin/zipkin-ruby
-    condition: "$TRAVIS_RUBY_VERSION == 2.6"
+    condition: "$TRAVIS_RUBY_VERSION == 2.7"
```
ykitamura-mdsol, line 28: Travis is trying to deploy on both gemfiles...
https://travis-ci.org/github/openzipkin/zipkin-ruby/jobs/675596484

PR #175, opened 2020-05-19
Add trace_context option to TraceWrapper to retrieve trace data

ykitamura-mdsol:
Added this option especially for the consumer side of messaging tracing to retrieve trace data easily.

The logic is a subset of the `ZipkinEnv` class:
https://github.com/openzipkin/zipkin-ruby/blob/0.44.0/lib/zipkin-tracer/rack/zipkin_env.rb#L39

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,6 +242,14 @@ TraceWrapper.wrap_in_custom_span(config, "custom span") do |span|
 end
 ```
 
+The `trace_context:` keyword argument can be used to retrieve trace data:
+```ruby
+trace_context = { trace_id: '234555b04cf7e099', span_id: '234555b04cf7e099' }
+
+TraceWrapper.wrap_in_custom_span(config, "custom span", trace_context: trace_context) do |span|
+  :
+end
+```
 
 ## Development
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -242,6 +242,14 @@ TraceWrapper.wrap_in_custom_span(config, "custom span") do |span|
 end
 ```
 
+The `trace_context:` keyword argument can be used to retrieve trace data:
+```ruby
+trace_context = { trace_id: '234555b04cf7e099', span_id: '234555b04cf7e099' }
+
+TraceWrapper.wrap_in_custom_span(config, "custom span", trace_context: trace_context) do |span|
```
jcarres-mdsol, line 249: Is trace_contexst going to grow a lot? Seems little bit weird that we create a hash just for two keys?

adriancole: seems it should carry the sampling status also right?

ykitamura-mdsol: I guess my example here is not good.. 😅 

These are the minimum required keys and in reality, you are going to pass more keys like:
```ruby
trace_context = { 
  trace_id: '234555b04cf7e099', 
  span_id: '234555b04cf7e099', 
  sampled: 'true', 
  flags: '0'
}
```

I will update this example.

adriancole: ok in this case I would not use flags rather debug as that's more semantically compatible vs a relic of X-B3

ykitamura-mdsol: removed `flags` in 5f15a03

PR #176, opened 2020-05-26
Add Amazon SQS tracer

### LINKED ISSUE #136: SQS integration

ykitamura-mdsol:
Added SQS tracer to trace SQS messages.
This tracer will aggressively add trace data to all SQS messages.

Closes #136 

### DIFF
--- a/lib/zipkin-tracer/sqs/zipkin-tracer.rb
+++ b/lib/zipkin-tracer/sqs/zipkin-tracer.rb
@@ -1,0 +1,57 @@
+module ZipkinTracer
+  # This module is designed to prepend to the SQS client to add trace data as message attributes.
+  # https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-sqs/lib/aws-sdk-sqs/client.rb
+  module SqsHandler
+    def send_message(params = {}, options = {})
+      zipkin_sqs_trace_wrapper(params, __method__) { |params_with_trace| super(params_with_trace, options) }
+    end
+
+    def send_message_batch(params = {}, options = {})
+      zipkin_sqs_trace_wrapper(params, __method__) { |params_with_trace| super(params_with_trace, options) }
+    end
+
+    private
+
+    ZIPKIN_KEYS = %i[trace_id parent_id span_id sampled].freeze
+    ZIPKIN_REMOTE_ENDPOINT_SQS = Trace::Endpoint.remote_endpoint(nil, 'amazon-sqs')
+
+    def zipkin_sqs_trace_wrapper(params, method_name)
+      trace_id = TraceGenerator.new.next_trace_id
+      zipkin_set_message_attributes(params, method_name, trace_id)
+
+      TraceContainer.with_trace_id(trace_id) do
+        if Trace.tracer && trace_id.sampled?
+          Trace.tracer.with_new_span(trace_id, method_name) do |span|
+            span.kind = Trace::Span::Kind::PRODUCER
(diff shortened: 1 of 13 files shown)


lib/zipkin-tracer/sqs/zipkin-tracer.rb
```diff
@@ -0,0 +1,57 @@
...
+    def zipkin_sqs_trace_wrapper(params, method_name)
+      trace_id = TraceGenerator.new.next_trace_id
+      zipkin_set_message_attributes(params, method_name, trace_id)
+
+      TraceContainer.with_trace_id(trace_id) do
+        if Trace.tracer && trace_id.sampled?
+          Trace.tracer.with_new_span(trace_id, method_name) do |span|
+            span.kind = Trace::Span::Kind::PRODUCER
+            span.remote_endpoint = ZIPKIN_REMOTE_ENDPOINT_SQS
+            span.record_tag('queue.url', params[:queue_url])
```
ykitamura-mdsol, line 27: same as brave

https://github.com/openzipkin/zipkin-aws/blob/release-0.21.1/brave-instrumentation/aws-java-sdk-sqs/src/main/java/brave/instrumentation/aws/sqs/SendMessageTracingRequestHandler.java#L111-L112

lib/zipkin-tracer/zipkin_sender_base.rb
```diff
@@ -40,7 +39,10 @@ def start_span(trace_id, name, timestamp = Time.now)
 
     def skip_flush?(span)
       return true if span.kind == Trace::Span::Kind::CLIENT && span.has_parent_span?
-      return true if span.kind == Trace::Span::Kind::PRODUCER && spans.any? { |s| s.kind == Trace::Span::Kind::SERVER }
+
+      if span.kind == Trace::Span::Kind::PRODUCER
+        return true if spans.any? { |s| s.kind == Trace::Span::Kind::SERVER || s.kind == Trace::Span::Kind::CONSUMER }
```
ykitamura-mdsol, line 44: include CONSUMER span kind to support chaining multiple queues

jcarres-mdsol:
Many required tags in there. Some may be difficult to fill with proper data

adriancole:
^^ required is work in progress. if something doesn't exist, we'd just flip that to optional as it makes no sense to make something required that can't be derived. any specific ones catch your eye?

ps why required at all? we didn't make things like http.method required in the past and for lack of making anything required it can make things like span names difficult. probably need at least one required tag, at least required to be able to parse even if not actually tagged.

jcarres-mdsol:
In this PR we would be overwriting the AWS SDK to instrument any call to SQS and add tracing previous to the call.
I like this because importing the class will auto-instrument your code, zero effort.
But we lose the chance of adding extra information to those spans. So whatever tag added is something that we have access to from within the AWS SDK, which may not be a lot. I'll wait and see what can Yohei find here

ykitamura-mdsol:
I'm a little confused about what needs to be done at this point...
Maybe revisit after the java side is done?

adriancole:
sorry it was just a note that we should either wait or revisit later. either is fine. I just didn't want you surprised later

ykitamura-mdsol:
No worries! Thanks for the heads up!

PR #177, opened 2020-08-27
Add a 'check_routes' option to make the routable request check optional

ykitamura-mdsol:
### Background

The Rails framework uses a hidden `_method` parameter to send "PATCH", "PUT", and "DELETE" requests over "POST":
https://guides.rubyonrails.org/form_helpers.html#how-do-forms-with-patch-put-or-delete-methods-work-questionmark

If "POST" is excluded in the `config/routes.rb` file like this:
```ruby
resources :users, except: [:create] do
  :
```

"PATCH", "PUT", and "DELETE" requests are also not recognizable by this method:
`Rails.application.routes.recognize_path(path_info, method: http_method)`

add the `Application.routable_request?` method will return `false`:
https://github.com/openzipkin/zipkin-ruby/blob/938f9a3fe7c88e75ad8fd0ac6a08bf9d4a784773/lib/zipkin-tracer/application.rb#L4-L13

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ use ZipkinTracer::RackHandler, config
 * `:service_name` **REQUIRED** - the name of the service being traced. There are two ways to configure this value. Either write the service name in the config file or set the "DOMAIN" environment variable (e.g. 'test-service.example.com' or 'test-service'). The environment variable takes precedence over the config file value.
 * `:sample_rate` (default: 0.1) - the ratio of requests to sample, from 0 to 1
 * `:sampled_as_boolean` - When set to true (default but deprecrated), it uses true/false for the `X-B3-Sampled` header. When set to false uses 1/0 which is preferred.
+* `:check_routes` - When set to true, only routable requests are sampled.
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -21,6 +21,7 @@ use ZipkinTracer::RackHandler, config
 * `:service_name` **REQUIRED** - the name of the service being traced. There are two ways to configure this value. Either write the service name in the config file or set the "DOMAIN" environment variable (e.g. 'test-service.example.com' or 'test-service'). The environment variable takes precedence over the config file value.
 * `:sample_rate` (default: 0.1) - the ratio of requests to sample, from 0 to 1
 * `:sampled_as_boolean` - When set to true (default but deprecrated), it uses true/false for the `X-B3-Sampled` header. When set to false uses 1/0 which is preferred.
+* `:check_routes` - When set to true, only routable requests are sampled.
```
jfeltesse-mdsol, line 24: explictily mention that it's false by default?

ykitamura-mdsol: added b86b191

PR #178, opened 2020-09-23
Fix to set SERVER span kind at the beginning to avoid being flushed before closing

ykitamura-mdsol:
If a `SERVER` span has a `PRODUCER` child span, it will be flushed when the child span is closed if span kind is not yet set:

https://github.com/openzipkin/zipkin-ruby/blob/2c71f1dccd5df9a0dcbb0358b9d78b6ca0fa1270/lib/zipkin-tracer/zipkin_sender_base.rb#L43-L45

@adriancole @jcarres-mdsol

### DIFF
--- a/spec/lib/rack/zipkin-tracer_spec.rb
+++ b/spec/lib/rack/zipkin-tracer_spec.rb
@@ -215,4 +215,23 @@ describe ZipkinTracer::RackHandler do
 
     end
   end
+
+  context 'SERVER span kind' do
+    subject { middleware(app) }
+
+    let(:span) { double }
+
+    before do
+      allow(tracer).to receive(:start_span).and_return(span)
+    end
+
+    it 'sets SERVER span kind before calling trace!' do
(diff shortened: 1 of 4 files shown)


jfeltesse-mdsol:
Would it be possible to add minimal specs to ensure we don't run into this again in the future?

ykitamura-mdsol:
made slight changes and added a spec in 780ef23

jfeltesse-mdsol (approved):

spec/lib/rack/zipkin-tracer_spec.rb
```diff
@@ -215,4 +215,23 @@ def expect_tags(path = '/')
...
+  context 'SERVER span kind' do
+    subject { middleware(app) }
+
+    let(:span) { double }
+
+    before do
+      allow(tracer).to receive(:start_span).and_return(span)
+    end
+
+    it 'sets SERVER span kind before calling trace!' do
```
adriancole, line 228: good

adriancole (approved):

PR #179, opened 2020-10-12
Fix to not flush local component span when a server span or a consumer span is in the stack

ykitamura-mdsol:
follow up on #178

No span kind is specified for local component spans: 
https://github.com/openzipkin/zipkin-ruby/blob/50640d9c852c77c33670cedf61031b04b46ea984/lib/zipkin-tracer/trace_client.rb#L20-L23

Updated to not flush local component span (span.kind = `nil`) when SERVER or CONSUMER span is in the stack.

@adriancole @jcarres-mdsol

### DIFF
--- a/lib/zipkin-tracer/zipkin_sender_base.rb
+++ b/lib/zipkin-tracer/zipkin_sender_base.rb
@@ -40,7 +40,7 @@ module Trace
     def skip_flush?(span)
       return true if span.kind == Trace::Span::Kind::CLIENT && span.has_parent_span?
 
-      if span.kind == Trace::Span::Kind::PRODUCER
+      if span.kind.nil? || span.kind == Trace::Span::Kind::PRODUCER
         return true if spans.any? { |s| s.kind == Trace::Span::Kind::SERVER || s.kind == Trace::Span::Kind::CONSUMER }
       end
     end
--- a/lib/zipkin-tracer/zipkin_sender_base.rb
+++ b/lib/zipkin-tracer/zipkin_sender_base.rb
@@ -38,11 +38,9 @@ module Trace
     end
 
     def skip_flush?(span)
-      return true if span.kind == Trace::Span::Kind::CLIENT && span.has_parent_span?
+      return false if span.kind == Trace::Span::Kind::SERVER || span.kind == Trace::Span::Kind::CONSUMER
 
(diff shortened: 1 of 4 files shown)


lib/zipkin-tracer/zipkin_sender_base.rb
```diff
@@ -40,7 +40,7 @@ def start_span(trace_id, name, timestamp = Time.now)
     def skip_flush?(span)
       return true if span.kind == Trace::Span::Kind::CLIENT && span.has_parent_span?
 
-      if span.kind == Trace::Span::Kind::PRODUCER
+      if span.kind.nil? || span.kind == Trace::Span::Kind::PRODUCER
         return true if spans.any? { |s| s.kind == Trace::Span::Kind::SERVER || s.kind == Trace::Span::Kind::CONSUMER }
```
jcarres-mdsol, line 44: I wonder, should we rewrite this method to check return true if any span in spans except the current is s.kind == Trace::Span::Kind::SERVER || s.kind == Trace::Span::Kind::CONSUMER

Wouldn't that be sufficient?

mdeemoe:
@jcarres-mdsol @ykitamura-mdsol 
Is there an ETA on this and on the corresponding Checkmate PR needed? 

This is holding up deploying the release to distro and validation. We can't wait much longer before we have to move forward without this fix. We are planning to send Checkmate to Validation tomorrow.

lib/zipkin-tracer/zipkin_sender_base.rb
ykitamura-mdsol: how about
```ruby
    def skip_flush?(span)
      case span.kind
      when Trace::Span::Kind::SERVER, Trace::Span::Kind::CONSUMER
        false
      when Trace::Span::Kind::CLIENT
        span.has_parent_span?
      else
        spans.any? { |s| s.kind == Trace::Span::Kind::SERVER || s.kind == Trace::Span::Kind::CONSUMER }
      end
    end
```
?

ykitamura-mdsol: or you mean we don't even need to check `span.has_parent_span?` ?

ykitamura-mdsol: updated in ba42be2

jcarres-mdsol (approved):

PR #180, opened 2020-11-26
Moves CI to GitHub Actions

adriancole:
This moves from Travis to GitHub Actions based test and deploy workflow.

Currently, this does not include testing on JRuby, but that could be added in the future.

Notably, this uses standard script conventions used by other projects in OpenZipkin. The main change to the actual code here is reverting  version.rb to 0.0.0 like some other Ruby projects are doing, instead of updating it manually in git each time.

## Release overview

Primarily what this does is add `build-bin/configure_test`, `build-bin/test`, `build-bin/configure_deploy` and `build-bin/deploy`. While documented in the build-bin/README.md, the TL;DR; is that pushing a tag like 0.5.0 will result in a ruby gems push. I've already setup my api key for this in openzipkin org secrets, so the next release should be fine.

## Motivation

travis-ci.org is being shut down the end of the year. We tried switching to travis-ci.com, but shortly ran out of quota. For the last two weeks we have not succeeded in getting classified as OSS or even a temporary quota increase. In response to this, all the other projects have moved, or are in the process of moving to GitHub Actions.

While at it, we are standardizing build script and job names for easy reference when people switch from one project to another. It is important that the tasks done in CI do not heavily rely on CI specific tools as in the future we may get into trouble like this again. For example, in the past we had to switch from travis to circle back to travis. Using scripts we control as entrypoints gives us flexibility as CI providers change their policies around open source communities.


.github/ISSUE_TEMPLATE/feature.md
```diff
@@ -0,0 +1,19 @@
+---
+name: Feature Request
+about: Please first, look at existing issues to see if the feature has been requested before.
+labels: enhancement
+---
+Please first, look at [existing issues](https://github.com/openzipkin/brave-karaf/issues) to see if the feature has been requested before. If you don't find anything tell us what problem you’re trying to solve. Often a solution already exists! Don’t send pull requests to implement new features without first getting our support. Sometimes we leave features out on purpose to keep the project small.
```
adriancole, line 6:
```suggestion
Please first, look at [existing issues](https://github.com/openzipkin/zipkin-ruby/issues) to see if the feature has been requested before. If you don't find anything tell us what problem you’re trying to solve. Often a solution already exists! Don’t send pull requests to implement new features without first getting our support. Sometimes we leave features out on purpose to keep the project small.
```

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,33 @@
+# Contributing to Zipkin
+
+If you would like to contribute code, fork this GitHub repository and
+send a pull request (on a branch other than `master` or `gh-pages`).
+
+When submitting code, please apply [Square Code Style](https://github.com/square/java-code-styles).
```
jcchavezs, line 6: Does this remain true?

adriancole: nope copy/pasta!

jfeltesse-mdsol:
https://blog.travis-ci.com/oss-announcement says 

> Open source accounts, as always, will be completely free under travis-ci.com.

Doesn't it apply to openzipkin as well?

adriancole:
travis have literally not granted openzipkin OSS status. I applied for this in both docker and travis, neither approved us, nor given us any reason why not after 2 weeks.

adriancole:
FYI it is still the case that neither travis nor docker have approved us as OSS. Travis did give some extra credits meanwhile, which is helpful.

That said, this change should still be merged imho. It is better documented and more comprehensive to the former.

It is fine to also add travis additionally on top of this, ex if we want to have multiple checks, but we can't have a dev zero situation either. Hence, I maintain the release process shouldn't rely on a rate limited CI. Make sense?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,33 @@
...
+    runs-on: ubuntu-20.04  # newest available distribution, aka focal
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 1  # full history is not needed
+      - name: Set up Ruby
+        uses: actions/setup-ruby@v1
+        with:
+          ruby-version: 2.6.x
```
anuraaga, line 24:
```suggestion
          ruby-version: 2.7.x
```

Seems like it repros what we have in https://github.com/openzipkin/zipkin-ruby/blob/master/.travis.yml#L28

anuraaga (approved):

lib/zipkin-tracer/version.rb
```diff
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
 module ZipkinTracer
-  VERSION = '0.47.2'
+  # Replaced with real version during CI as a part of /build-bin/gem/gem_push
+  VERSION = '0.0.0'
```
jcchavezs, line 5: I'd rather do `VERSION = 'dev'` instead as I believe that is the tracer version we will see in local?

adriancole: cool I have no opinion on that. yeah this would be the version string for a gemfile when trying to use this published locally

adriancole:
```suggestion
  VERSION = 'dev'
```

jcchavezs (approved):

lib/zipkin-tracer/version.rb
adriancole: sorry but ruby is not ok with a version like "dev" I need to roll this back

lib/zipkin-tracer/version.rb
```diff
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
 module ZipkinTracer
-  VERSION = '0.47.2'
+  # Replaced with real version during CI as a part of /build-bin/gem/gem_push
+  VERSION = 'dev'
```
adriancole, line 5:
```suggestion
  VERSION = '0.0.0'
```

.github/ISSUE_TEMPLATE/feature.md
```diff
@@ -0,0 +1,19 @@
+---
+name: Feature Request
+about: Please first, look at existing issues to see if the feature has been requested before.
+labels: enhancement
+---
+Please first, look at [existing issues](https://github.com/openzipkin/zipkin-ruby/issues) to see if the feature has been requested before. If you don't find anything tell us what problem you’re trying to solve. Often a solution already exists! Don’t send pull requests to implement new features without first getting our support. Sometimes we leave features out on purpose to keep the project small.
+
+**Feature:**
```
jfeltesse-mdsol, line 8: how about having the same formatting as in the bug.md? e.g. use markdown headers instead of bold.

.github/ISSUE_TEMPLATE/feature.md
```diff
@@ -0,0 +1,19 @@
...
+Description of the feature
+
+**Rational**
```
jfeltesse-mdsol, line 11: Rational*e*

.github/ISSUE_TEMPLATE/feature.md
```diff
@@ -0,0 +1,19 @@
...
+Why would this feature help others besides me?
+
+**Example Scenario**
+What use case is not practical to serve before this feature?
```
jfeltesse-mdsol, line 15: Reads a bit clunky IMHO, how about something like "What kind of use cases would benefit from this feature?"

.github/ISSUE_TEMPLATE/feature.md
```diff
@@ -0,0 +1,19 @@
...
+
+**Prior Art**
+ * Links to prior art
```
jfeltesse-mdsol, line 18: [nits] no need to have a space at col 1

.github/workflows/test.yml
```diff
@@ -0,0 +1,40 @@
...
+    branches: master
+    paths-ignore: '**/*.md'
+
+jobs:
+  test:
+    runs-on: ubuntu-20.04  # newest available distribution, aka focal
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby: [2.5, 2.6, 2.7]
```
jfeltesse-mdsol, line 23: what about jruby?

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 0.47.3
+* Moved CI to [GitHub Actions](https://github.com/openzipkin/zipkin-ruby/workflows)
```
jfeltesse-mdsol, line 2: [nits] end the line with "."

RELEASE.md
```diff
@@ -1,6 +1,44 @@
...
+
+1. **Alert others you are releasing**
+
+   There should be no commits made to master while the release is in progress (about 10 minutes). Before you start
+   a release, alert others on [gitter](https://gitter.im/openzipkin/zipkin) so that they don't accidentally merge
+   anything. If they do, and the build fails because of that, you'll have to recreate the release tag described below.
+
+1. **Push a git tag**
+
+   The tag should be of the format `N.M.L`, ex `git tag 1.18.1; git push origin 1.18.1`.
```
jfeltesse-mdsol, line 13: 1) I don't know what N.M.L stands for but I think it'd be clearer to use semver terminology, e.g. `MAJOR.MINOR.PATCH`

2) The tag commands should be separated by `&&` instead of `;` so it doesn't try to push if the local tag command failed.

RELEASE.md
```diff
@@ -1,6 +1,44 @@
...
+
+1. **Wait for CI**
+
+   The `N.M.L` tag triggers [`build-bin/deploy`](build-bin/deploy), which does the following:
+     * https://rubygems.org/gems/zipkin-tracer [`build-bin/gem/gem_push`](build-bin/gem/gem_push)
+
+## Credentials
+
+The release process uses various credentials. If you notice something failing due to unauthorized,
+look at the notes in [.github/workflows/deploy.yml] and check the [org secrets](https://github.com/organizations/openzipkin/settings/secrets/actions)
```
jfeltesse-mdsol, line 23: [nits] missing "." at the end

build-bin/README.md
```diff
@@ -0,0 +1,203 @@
...
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0  # full git history
+      - name: Test
+        run: build-bin/configure_test && build-bin/test
+```
+
+### Example Travis setup
```
jfeltesse-mdsol, line 82: remove travis entirely so we don't need people to know about _both_ GA and travis? I saw your comment about the additional credits but it doesn't sound like it's worth bothering any longer.

adriancole:
@jfeltesse-mdsol excellent comments, you caught some things not yet noticed. I will address them.

WRT removing travis from the README, I think it is still important. One thing I found the last 5 years is that CI providers take turns being the one that is broken. Maybe next year GH actions cut us off and we have to use another provider. It took me almost an entire fulltime month to figure out all this stuff. Intentionally deleting docs will only sacrifice someone else to such a fate, and I think few will have time for it.

Having a couple different examples are helpful as features are not uniform between them. For example, I have recently setup secondary tests in the docker-alpine project that can only be done in travis as it has arm64 nodes, something no other provider has. If those jobs fail to run, we are still ok because emulation tests exist. However, it is better to run tests when we can vs never. The same could be said about JRuby, make sense?

adriancole:
ex here's testing things that are nice to test, but won't kill us if they skip.. programmatic dependency on travis https://github.com/openzipkin/docker-alpine/blob/master/.travis.yml

build-bin/README.md
adriancole: I can see this point for sure, mentioned in top-level issue comments by accident sorry! TL;DR; is that we can have a soft dep on travis in some cases where it helps. literally this file is the same for all the other repos, so it is nice to take advantage of commonality where used ex here https://github.com/openzipkin/docker-alpine/blob/master/.travis.yml

jfeltesse-mdsol: Makes sense, thanks for the thorough reply!

.github/ISSUE_TEMPLATE/feature.md
adriancole: agree it is awkward, I was trying to rephrase brain into considering if it is actually possible now or not, but instead, just broke the brain!

adriancole:
ok I found I was using the wrong setup-ruby. now there's 100pct coverage of the matrix travis had. I might recommend doing boundaries (2.3, 2.7) instead of every minor between 2.3 and 2.7, unless you know a reason otherwise. It will keep the queuing down which isn't hugely important but nicer I guess.

Thanks for the feedback and I'll propagate it to all the other repos!

adriancole:
I don't like setting traps for people so I deployed 0.47.3 via the tag of the same name

https://github.com/openzipkin/zipkin-ruby/tree/0.47.3

If for some reason this fails, I'll clean it up

adriancole:
Screen Shot 2020-12-04 at 2 39 03 PM

https://github.com/openzipkin/zipkin-ruby-example

a-ok (after some fuzz in the commit history, I'll admit)

Issue #125, opened 2018-07-11
Add setting to disable shared spans (joins) in tracer. 


tylerbenson:
Other tracer implementations include a setting to force client and server spans to have different spanId's.  In Go the setting is [`WithSharedSpans(false)`](https://godoc.org/github.com/openzipkin/zipkin-go#WithSharedSpans) and in Java it's [`supportsJoin(false)`](https://static.javadoc.io/io.zipkin.brave/brave/5.0.0/brave/Tracing.Builder.html#supportsJoin-boolean-).

There should be something similar in ruby.

This is important because zipkin traces may be reported to non-zipkin backends that might not support the concept of joining spans.

dp9463:
agree. I encountered a problem when use jaeger client send request to zipkin client, and the backend server is jaeger server. It will report the duplicate span ID.

PR #182, opened 2023-01-07
Add setting to disable shared spans (joins) in tracer

timward60:
Addresses #125 by adding setting to disable support for joins.

When set to false, the it will force client and server spans to have different spanId's. This may be needed because zipkin traces may be reported to non-zipkin backends that might not support the concept of joining spans.


timward60:
@jcchavezs You were involved with a discussions from the issue. If you have some time can you help me with this review?

